PR #191, opened 2026-08-24
secrets: flag ASIA-prefixed temporary AWS key IDs

### LINKED ISSUE #167: AWS key detection misses ASIA-prefixed (STS) credentials
## Problem

`internal/packages/secrets.go:48`: `\bAKIA[0-9A-Z]{16}\b` matches only long-lived access key IDs, not `ASIA`-prefixed temporary/session credentials from STS/assumed roles — extremely common in modern AWS setups.

## Fix direction

Extend the pattern to `\b(AKIA|ASIA)[0-9A-Z]{16}\b`.

## Confidence

High on the gap; this is within the accepted "documented list, not exhaustive" tradeoff of ADR 0009, but an easy, obvious extension.

MsfPablo:
## Summary

The AWS access key ID pattern in `internal/packages/secrets.go` matched only `AKIA` (long-lived user keys), so `ASIA`-prefixed temporary STS/assumed-role session credentials passed the scan unflagged. Both share the same 16-character suffix shape, so this widens the existing pattern to `\b(AKIA|ASIA)[0-9A-Z]{16}\b`.

## Linked Issue

Closes #167

## Package Impact

## Safety

Findings still carry only the path and reason, never the matched value — `TestScanForSecretsNeverIncludesMatchedValue` continues to cover that.

## Verification

Relevant test cases:

- `TestScanForSecretsFlagsAWSSessionKeyID` (new) — an `ASIA`-prefixed key in `references/session.txt` produces a finding.
- `TestScanForSecretsFlagsAWSKeyID` — unchanged, `AKIA` still matches.
- `TestScanForSecretsIgnoresSafeContent` — unchanged, no new false positives.

```text
go build ./...
go test ./internal/packages/
go test ./...
ok  	github.com/agentic-lineage/lineage/internal/packages	0.645s
```

## Notes For Reviewers

The new test fixture splits the literal string (`"ASIA" + "ABCDEF..."`) the same way the existing `AKIA` fixtures do, to keep push protection from flagging the test file itself.

### DIFF
--- a/internal/packages/secrets.go
+++ b/internal/packages/secrets.go
@@ -49,7 +49,9 @@ var deniedExtensions = []string{".pem", ".key", ".pfx", ".p12"}
 // are essentially never present in legitimate package source material.
 var secretContentPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`-----BEGIN [A-Z ]*PRIVATE KEY-----`),
-	regexp.MustCompile(`\bAKIA[0-9A-Z]{16}\b`),         // AWS access key ID
+	// AWS access key ID: AKIA is a long-lived user key, ASIA a temporary
+	// STS/assumed-role session key.
+	regexp.MustCompile(`\b(AKIA|ASIA)[0-9A-Z]{16}\b`),
 	regexp.MustCompile(`\bgh[pousr]_[A-Za-z0-9]{20,}`), // GitHub token prefixes (ghp_, gho_, ghu_, ghs_, ghr_)
 }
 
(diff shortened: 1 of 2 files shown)


priyam-jain-2002 (commented):
Code review looks good, and I verified go test ./... locally. The GitHub Test workflow is green; the remaining PR Policy failure is compliance-only: the Verification block needs to include the literal `go test ./...` command because this PR adds test coverage. Please update the PR body with that full-suite command so PR Policy can pass, then this should be ready to approve and merge.

priyam-jain-2002 (approved):
Approved. The ASIA prefix is the missing sibling to AKIA for temporary AWS credentials, and the regression test keeps the matched value out of the finding while covering the new pattern. PR Policy and Test are green after the maintainer compliance-only body edit and branch update; I also verified go test ./... locally.

PR #194, opened 2026-08-24
cli: normalize the ref in disable the way enable does

### LINKED ISSUE #162: disable doesn't normalize package refs the way enable does
## Problem

`enableRef` re-expresses a `./`-relative ref against the project root when invoked from a subdirectory (verified by `TestEnableFromSubdirectoryUsesProjectRoot`), storing e.g. `./packages/sub/local-pack`. `runDisable` (`internal/cli/cli.go:830-853`) has no equivalent normalization — it does an exact-string `contains()` check against the literal CLI argument.

## Scenario

From `project/packages/sub`, `lineage enable ./local-pack` stores `./packages/sub/local-pack`. Running `lineage disable ./local-pack` from the same directory fails with `package "./local-pack" is not enabled`, even though it's the exact ref string `enable` accepted moments earlier.

## Fix direction

Reuse the same ref-normalization helper `enableRef` already has before comparing against `EnabledPackages`.

## Confidence

High — confirmed against source and existing test coverage for the equivalent `enable` behavior.

MsfPablo:
## Summary

`enableRef` re-expresses a `./`-relative ref against the project root before storing it, so `lineage enable ./local-pack` from `project/packages/sub` writes `./packages/sub/local-pack`. `runDisable` compared the literal CLI argument against `enabled_packages`, so `lineage disable ./local-pack` from that same directory failed with `package "./local-pack" is not enabled` — rejecting the exact ref enable had accepted moments earlier.

This lifts the normalization enable was doing inline into `projectRelativeRef` and calls it from both sides, so the two commands agree on what string a given ref means.

If the ref no longer resolves — the package directory being gone is a common reason to be disabling it — disable falls back to the literal argument, which still exactly matches anything enable stored verbatim. The error message keeps quoting what the user typed, not the normalized form.

## Linked Issue

Closes #162

## Package Impact

## Safety

Resolution still goes through `packages.ResolveReference`; this only changes which string is compared against config, and only ever narrows to a ref that already resolves.

## Verification

Relevant test cases:

- `TestDisableFromSubdirectoryMatchesStoredRef` (new) — enable then disable `./local-pack` from a subdirectory; the round trip now succeeds and leaves `enabled_packages` empty. Confirmed it fails on `develop` with exactly the reported error before the fix.
- `TestEnableFromSubdirectoryUsesProjectRoot` — unchanged, still passes over the extracted helper.

```text
go test ./...
ok  	github.com/agentic-lineage/lineage/internal/cli	3.753s
(all packages ok)
```

## Notes For Reviewers

The extraction is behavior-preserving for enable: `projectRelativeRef` folds the old `strings.HasPrefix(ref, ".") && resolveRoot != projectRoot` guard and the `filepath.Rel` failure fallback in unchanged.

Disable resolves the ref once to normalize it, which enable already had in hand. I kept the resolve local rather than restructuring `runDisable`'s config lookup, to keep the diff to the matching bug.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -558,6 +558,28 @@ func runEnable(args []string, home string, stdin *bufio.Reader, stdout, stderr i
 // left the workspace unchanged), but it is also not success - callers like
 // runAdd need to tell the two apart instead of treating "no error" as
 // "enabled" and reporting readiness for a package that was never turned on.
+// projectRelativeRef re-expresses a "./" or "../" style ref - typed
+// relative to cwd - against projectRoot, which is where the config lives
+// and what `lineage run` later resolves enabled_packages entries against.
+// Anything else, including a ref typed at the project root itself, is
+// already stored verbatim and comes back unchanged. Both enable (which
+// writes the entry) and disable (which has to match it) go through this,
+// so the two agree on what string a given ref means.
+func projectRelativeRef(ref, projectRoot, resolveRoot, resolvedPath string) string {
+	if !strings.HasPrefix(ref, ".") || resolveRoot == projectRoot {
+		return ref
+	}
+	rel, err := filepath.Rel(projectRoot, resolvedPath)
+	if err != nil {
+		return ref
+	}
+	rel = filepath.ToSlash(rel)
+	if !strings.HasPrefix(rel, ".") {
+		rel = "./" + rel
+	}
+	return rel
+}
+
 func enableRef(ref, home string, autoApprove bool, stdin *bufio.Reader, stdout, stderr io.Writer) (bool, error) {
 	cwd, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


priyam-jain-2002 (commented):
Code review looks good, and I verified go test ./... locally. The enable/disable normalization behavior is now shared through the helper, and the regression test covers the reported subdirectory round trip. I did not approve or merge this yet because after #196 merged, GitHub now reports this branch as CONFLICTING / DIRTY against develop. Please rebase or otherwise resolve the conflict, then let CI rerun; once it is mergeable again this should be ready for approval.

priyam-jain-2002 (approved):
Approved. This keeps enable and disable using the same project-relative ref normalization and preserves the fallback for refs that no longer resolve. The regression test covers the reported subdirectory round trip. PR Policy and Test are green after automatic branch update, and I verified go test ./internal/cli plus go test ./... locally.

PR #195, opened 2026-08-24
auth: re-enforce 0600 on an existing token file

### LINKED ISSUE #159: GitHub token file permissions aren't re-enforced when the token file already exists
## Problem

`SaveToken` (`internal/auth/token.go:19-27`) writes with `os.WriteFile(TokenPath(home), ..., 0o600)`. Per POSIX `open()` semantics, the requested mode only applies when the file is *created*; if `github_token` already exists (older lineage version, manual edit, restored backup), its current permission bits are left untouched. A stale or tampered token file with loose permissions stays loose across every subsequent `lineage login`.

## Fix direction

`os.Chmod(path, 0o600)` explicitly after the write, regardless of whether the file existed before.

## Confidence

Medium — mechanism confirmed against POSIX semantics and source; `TestSaveLoadDeleteTokenRoundTrip` only checks perms on a freshly-created file.

MsfPablo:
## Summary

`SaveToken` relied on the mode argument to `os.WriteFile`, which POSIX `open()` applies only when the file is created. A `github_token` that already existed — written by an older Lineage, edited by hand, restored from a backup — kept whatever permissions it already had, so a world-readable bearer credential stayed world-readable across every subsequent `lineage login`.

This adds an explicit `os.Chmod` after the write, and names the mode so the create path and the re-enforce path cannot drift apart.

## Linked Issue

Closes #159

## Package Impact

## Safety

This narrows who can read a stored credential; nothing new is written or exported. A `chmod` failure is surfaced as an error rather than swallowed, so login cannot quietly report success while leaving the token loose.

## Verification

Relevant test cases:

- `TestSaveTokenTightensExistingLoosePermissions` (new) — pre-creates `github_token` at `0644`, saves over it, asserts no group/other bits survive. Fails on `develop` with `token file mode = 644`. Skipped on Windows, matching the existing permission assertion.
- `TestSaveLoadDeleteTokenRoundTrip` — unchanged; it only covered the freshly-created case, which is why this slipped through.

```text
go test ./internal/auth/
go test ./...
ok  	github.com/agentic-lineage/lineage/internal/auth	5.598s
```

## Notes For Reviewers

`os.Chmod` is a no-op cost on the common path (file just created at the right mode), so I did not gate it behind a stat-and-compare.

Worth noting the scope limit: this fixes permissions going forward from the next `lineage login`. It does not repair a loose token file that exists right now and is never saved over again. If you want that covered, `LoadToken` could tighten on read too — happy to file or fold that in, but it changes a read path into a write path, so I left it out of this fix.

### DIFF
--- a/internal/auth/token.go
+++ b/internal/auth/token.go
@@ -7,6 +7,10 @@ import (
 	"strings"
 )
 
+// tokenFileMode keeps the stored bearer credential readable only by its
+// owner.
+const tokenFileMode = 0o600
+
 // TokenPath is where `lineage login` stores the GitHub device-flow token.
 func TokenPath(home string) string {
 	return filepath.Join(home, "github_token")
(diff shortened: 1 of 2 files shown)


priyam-jain-2002 (commented):
Code review looks good, and I verified go test ./... locally. The chmod-after-write behavior correctly handles existing loose token files, and surfacing chmod errors keeps login from reporting success while leaving credentials readable. GitHub Test is green; the remaining PR Policy failure is compliance-only: the Verification block needs to include the literal `go test ./...` command because this PR adds test coverage. Please update the PR body with that full-suite command so PR Policy can pass, then this should be ready to approve and merge.

priyam-jain-2002 (approved):
Approved. Re-applying 0600 after WriteFile fixes the existing-file permission gap, and surfacing chmod failures keeps credential storage from silently succeeding in an unsafe state. PR Policy and Test are green after the maintainer compliance-only body edit and branch update; I also verified go test ./... locally.

PR #196, opened 2026-08-24
cli: validate the workspace name before joining it into the home dir

### LINKED ISSUE #171: lineage init workspace allows path traversal outside ~/.lineage/workspaces
## Problem

`runInit` (`internal/cli/cli.go:118`): `dir := config.WorkspacePackagesDir(home, args[1])` → `filepath.Join(home, "workspaces", name, "packages")` — no validation of `name`.

## Scenario

Reproduced: `lineage init workspace ../../escaped` creates `.../escaped/packages` outside the intended `workspaces/` directory.

## Fix direction

Validate the workspace name against a safe charset before joining, same fix shape as the manifest-name validation issue already filed.

## Confidence

High — confirmed by direct reproduction. Lower severity since it's a local, self-directed action, but Lineage's docs promote copy/paste bootstrap prompts run by an agent, so a bad or attacker-influenced workspace name is a plausible vector.

MsfPablo:
## Summary

`lineage init workspace <name>` passed the name straight to `config.WorkspacePackagesDir`, i.e. into `filepath.Join`, with no validation. `lineage init workspace ../../escaped` therefore created `.../escaped/packages` outside the intended `workspaces/` directory — reproduced, and now covered by a test that fails on `develop`.

A workspace name is a plain identifier, not a path, so this validates it against a safe charset up front rather than trying to sanitize the joined result.

## Linked Issue

Closes #171

## Package Impact

## Safety

The rejected charset also covers the Windows separator and leading dots, so a name that validates is usable as a single directory component on every supported platform, not only one where `/` happens to be the separator.

## Verification

Relevant test cases:

- `TestInitWorkspaceRejectsTraversingName` (new, `internal/cli`) — the reported `../../escaped` reproduction is rejected, no directory appears outside `WorkspacesDir`, and a well-formed name still initializes. Fails on `develop` with `error = nil, want a rejection`.
- `TestValidateWorkspaceName` (new, `internal/config`) — accept/reject table including `..`, `.`, `a/b`, `a\\b`, `.hidden`, and empty.
- `TestValidateWorkspaceNameKeepsJoinInsideWorkspacesDir` (new) — asserts the property the charset exists to protect, so the test does not just restate the regex.

```text
go test ./...
(all packages ok)
```

## Notes For Reviewers

I put `ValidateWorkspaceName` in `config` next to `WorkspacePackagesDir` rather than inline in `runInit`, so the check lives beside the join it protects and is available if another caller ever takes a workspace name. `runInit` is the only caller of `WorkspacePackagesDir` today.

Related but deliberately out of scope: `ResolveReference` builds `home/workspaces/<workspace>/packages` from `cfg.Workspace` (`internal/packages/discovery.go:263`) without going through `WorkspacePackagesDir`, so a hand-edited `workspace:` value in a project config is a separate read-side path worth its own issue. Happy to file it if you agree it's a real gap.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -130,6 +130,10 @@ func runInit(args []string, home string, stdout, stderr io.Writer) error {
 			fmt.Fprintln(stderr, err)
 			return err
 		}
+		if err := config.ValidateWorkspaceName(args[1]); err != nil {
+			fmt.Fprintln(stderr, err)
+			return err
+		}
 		dir := config.WorkspacePackagesDir(home, args[1])
 		if err := os.MkdirAll(dir, 0o755); err != nil {
 			fmt.Fprintln(stderr, err)
--- a/internal/cli/regression_test.go
+++ b/internal/cli/regression_test.go
@@ -263,3 +263,36 @@ func TestParseRunArgs(t *testing.T) {
 		})
 	}
 }
+
+// TestInitWorkspaceRejectsTraversingName covers #171: `lineage init
+// workspace ../../escaped` created `.../escaped/packages` outside the
+// intended workspaces/ directory, because the name went straight into
+// filepath.Join unvalidated.
+func TestInitWorkspaceRejectsTraversingName(t *testing.T) {
+	tmp := t.TempDir()
+	home := filepath.Join(tmp, "home", "lineage")
+	if err := os.MkdirAll(home, 0o755); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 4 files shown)


priyam-jain-2002 (approved):
Approved. This fixes the traversal at the right boundary by validating workspace names before they reach WorkspacePackagesDir, and the regression coverage checks both the rejected traversal case and a valid workspace path. I also verified go test ./... locally in addition to the green GitHub Test and PR Policy checks.

PR #197, opened 2026-08-24
config: nil-guard EnabledPackages after unmarshal

### LINKED ISSUE #168: EnabledPackages isn't nil-guarded after loading project config
## Problem

`LoadProjectConfig` (`internal/config/config.go:74-91`) explicitly re-initializes `ProviderPreferences` and `Providers` to empty maps if `nil` after `yaml.Unmarshal`, but `EnabledPackages` gets no equivalent treatment. A config file with an explicit `enabled_packages:` (empty/null) rather than an omitted key leaves `cfg.EnabledPackages` as `nil` instead of the documented `[]string{}` default.

## Fix direction

Add the same nil-guard already applied to the two map fields.

## Confidence

Medium — low practical impact (nil slices behave like empty ones for range/len/append) but inconsistent with the pattern already established for sibling fields.

MsfPablo:
## Summary

`LoadProjectConfig` re-initializes `ProviderPreferences` and `Providers` to empty collections when `yaml.Unmarshal` leaves them nil, but `EnabledPackages` got no equivalent treatment. A config with an explicit but valueless `enabled_packages:` key unmarshals to nil, overwriting the empty slice `DefaultProjectConfig` seeded, so the field came back nil instead of the documented `[]string{}`.

## Linked Issue

Closes #168

## Package Impact

## Safety

## Verification

Relevant test cases:

- `TestLoadProjectConfigNormalizesExplicitEmptyCollections` (new) — loads a config with all three keys present and valueless, asserts none of them come back nil. Fails on `develop` on the `EnabledPackages` assertion and passes the two sibling assertions, which is exactly the inconsistency this closes.

```text
go test ./...
(all packages ok)
```

## Notes For Reviewers

As the issue notes, practical impact is low — a nil slice ranges, lens, and appends like an empty one, so nothing was misbehaving. This is about the loaded config matching its documented shape, and about the next reader of `LoadProjectConfig` not having to work out why two of three collections are normalized. I kept the test covering all three fields so the guard can't be dropped from one of them again without a failure.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -88,6 +88,9 @@ func LoadProjectConfig(path string) (ProjectConfig, error) {
 	if cfg.Providers == nil {
 		cfg.Providers = map[string]Provider{}
 	}
+	if cfg.EnabledPackages == nil {
+		cfg.EnabledPackages = []string{}
+	}
 	return cfg, nil
 }
 
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -61,3 +61,28 @@ func TestFindProjectConfigMissing(t *testing.T) {
 		t.Fatal("FindProjectConfig() error = nil, want error")
 	}
 }
+
+// TestLoadProjectConfigNormalizesExplicitEmptyCollections covers #168: an
+// explicit `enabled_packages:` with no value unmarshals to nil, overwriting
+// the empty slice DefaultProjectConfig seeded. The two sibling map fields
+// were already guarded; this one was not.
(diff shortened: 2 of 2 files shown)


priyam-jain-2002 (approved):
Approved. This matches the existing LoadProjectConfig normalization pattern for the sibling collection fields, and the regression test covers the explicit empty enabled_packages case without broadening the change. GitHub PR Policy and Test are green, and I also verified go test ./... locally both on the PR branch and after combining it with the updated develop base.

PR #198, opened 2026-08-24
Fix help flag parsing and add regression tests

### LINKED ISSUE #172: -h/--help is only recognized as the literal first argument
## Problem

Every `isHelpFlag(args[0])` call site in `internal/cli/cli.go` only checks the first argument. Any flag preceding `--help` causes it to be swallowed as a positional argument instead of showing help.

## Scenario

Reproduced: `lineage enable --yes --help` → error `package "--help" not found in project, user, or workspace packages`, instead of enable usage text.

## Fix direction

Scan all args for `-h`/`--help` before falling through to normal argument parsing, rather than only checking position 0.

## Confidence

High — confirmed by direct reproduction.

AlessandroDiGioacchino:
## Summary

Changed files `internal/cli/cli.go` and `internal/cli/help_test.go`
I introduced function `hasHelpFlag` in the former and a couple of tests in the latter.
`hasHelpFlag(args)` checks if any string in `args` is a help flag.

## Linked Issue

Closes #172

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Safety

## Verification

Relevant test cases:

- `TestHelpFlagRecognizedAfterOtherArguments`
- `TestHasHelpFlagRequiresExactToken`

```text
go test ./internal/cli -run Help
go test ./...
go run ./cmd/lineage enable --yes --help
```

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/agentic-lineage/lineage/internal/auth"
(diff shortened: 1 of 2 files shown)


priyam-jain-2002 (approved):
Approved. The helper keeps help-flag handling exact while allowing help after other command arguments, and the regression tests cover both the desired late-help behavior and false positives like --helpful. I verified the targeted help tests, go test ./..., and the CLI smoke command locally after an automatic merge with current develop. PR Policy and Test are green after the maintainer compliance-only body edit and branch update.

PR #199, opened 2026-08-24
Design the .lineage artifact as a whole: contents, structure, and versioning

### LINKED ISSUE #59: Design the .lineage artifact as a whole: contents, structure, and versioning
## Problem

`.lineage` has grown into two distinct on-disk artifacts, each accreted one field at a time as separate phases landed, with no single place that defines what belongs in either one, what's safe to delete, or what stability guarantees it carries:

**Project-level `<project>/.lineage/`**
- `config.yaml` — enabled packages and provider binary overrides (`internal/config`)
- `materialized-<provider>.json` — per-provider idempotent materialization state (`internal/materialize`), one file per provider that has ever run

**User-level `~/.lineage/`**
- `user/packages/` — imported/authored packages
- `workspaces/<name>/packages/` — workspace-scoped packages
- `bin/` — generated provider shims (`internal/shim`)

None of this was designed as a unit. Each piece has reasonable behavior in isolation (materialization state is idempotent per ADR 0008, the manifest format is versioned per ADR 0005), but the container directories themselves have no equivalent spec:

- No schema/version field on the state files (`config.yaml`, `materialized-<provider>.json`), unlike `lineage.yaml`'s `schema` field for packages — a future format change has no clean way to signal itself.
- No stated policy on whether project-level `.lineage/` should be committed to the receiver's repo or gitignored by default. We already got burned once this session by an internal doc landing in the repo by accident; the same risk applies here if a receiver's `.lineage/` ends up committed with machine-local provider binary paths in `config.yaml`.
- No single doc enumerating what's safe for a user to delete and regenerate (materialization state, shims) versus what's authoritative and would lose data if removed (imported packages under `user/packages/`).
- `lineage doctor` (#52, open) currently has no defined scope for validating `.lineage`'s own structure/integrity, because that structure was never written down.

## Ask

Produce a deliberate design for `.lineage` as a first-class artifact — likely as a new entry in `docs/decisions/`, the way the package manifest format got ADR 0005. It should:

1. Enumerate every file/directory Lineage writes, split clearly between project-level `.lineage/` and user-level `~/.lineage/`.
2. Decide whether the project-level and/or user-level state files need their own schema-version field, mirroring `lineage.yaml`'s `schema` (ADR 0005), so old state can be detected and migrated instead of silently misread.
3. Decide the commit/gitignore convention for project-level `.lineage/` and document it (e.g. in `CONTRIBUTING.md` or a generated `.gitignore` entry from `lineage init`).
4. Classify each entry as regenerable (safe to delete, Lineage rebuilds it) vs. authoritative (deleting it loses user data), so `lineage doctor` and any future `lineage clean`-style command have a real spec to check against.
5. Note explicitly what's out of scope: this is about the container directory and its state files, not the package manifest format (ADR 0005) or materialization idempotency semantics (ADR 0008), both already decided.

## Non-goals

This is not about adding new features like content-addressed snapshots (#7) or lineage graph metadata (#6) — those are future capabilities that would *live inside* whatever this issue defines, not a reason to block on them.

## Acceptance Criteria

- A written spec (ADR) enumerating `.lineage` contents at both the project and user level.
- An explicit decision on state-file schema versioning.
- An explicit decision on the gitignore/commit convention, applied in `CONTRIBUTING.md` and/or `lineage init`.
- `lineage doctor`'s scope updated (or a follow-up filed) to validate against the new spec.

priyam-jain-2002:
## Summary

`.lineage` grew into a project-level (`<project>/.lineage/`) and user-level (`~/.lineage/`) artifact one field at a time across separate phases, with no single spec for what belongs where, what's schema-versioned, what's safe to delete, or whether it's meant to be committed. Adds ADR 0015 as that spec and closes the gaps it surfaces:

- `config.yaml` and `materialized-<provider>.json` now carry a `schema` field, defaulted for existing files and rejected if unsupported — matching `lineage.yaml`'s existing convention (ADR 0005). `graph.json` deliberately stays unversioned for now (bare array, not an object; the ADR documents why and what would trigger revisiting it).
- `lineage enable` now gitignores a project's `.lineage/` the first time it creates one — it can carry machine-local provider binary paths and regenerable cache, and should never land in a receiver's repo. Idempotent; doesn't fight a receiver who later removes the entry on purpose.
- The ADR enumerates every project- and user-level entry (including `graph.json`, the snapshot store, and `~/.lineage/github_token`, none of which existed when #59 was filed) and classifies each as regenerable or authoritative.
- `CONTRIBUTING.md`/`ROADMAP.md`/`CHANGELOG.md` updated to reflect the decision and point future contributors at the ADR.

## Linked Issue

Closes #59

- [x] This PR targets `develop`.

## Package Impact

- [x] Documentation only

## Safety

## Verification

Relevant test cases:

- `TestLoadProjectConfigDefaultsMissingSchemaToCurrent` / `TestLoadProjectConfigRejectsUnsupportedSchema` / `TestSaveProjectConfigRoundTripsSchema`
- `TestApplyWritesCurrentSchema` / `TestNeedsApprovalDefaultsMissingSchemaToCurrent` / `TestNeedsApprovalRejectsUnsupportedSchema`
- `TestEnsureGitignoredCreatesFile` / `TestEnsureGitignoredAppendsToExistingFile` / `TestEnsureGitignoredAppendsMissingTrailingNewline` / `TestEnsureGitignoredIsNoOpWhenAlreadyPresent` / `TestEnsureGitignoredRecognizesBareEntry`
- `TestEnableCreatesGitignoreEntryForNewProject` / `TestEnableTwiceDoesNotDuplicateGitignoreEntry` / `TestEnableDoesNotOverwriteReceiversRemovedGitignoreEntry`

```text
go build ./... && go vet ./... && go test ./...
```

## Notes For Reviewers

- PR #179 is open against the same function (`enableRef`) and reorders where `config.SaveProjectConfig` runs. This PR's gitignore hook is placed at the very end of the function's success path specifically so it stays correct regardless of which of these two merges first — flagging it in case a rebase is still needed.
- The ADR files a follow-up for extending `lineage doctor`'s scope (materialize-state schema/staleness, `graph.json` integrity, snapshot referential integrity) rather than building that validation speculatively in this PR — opening that issue separately.


docs/decisions/0015-the-lineage-directory-is-a-versioned-enumerated-container.md
```diff
@@ -0,0 +1,83 @@
...
+
+### 3. Commit / gitignore convention
+
+Project-level `<project>/.lineage/` must never be committed to a receiver's repo. `config.yaml` can carry machine-local provider binary paths (`providers.<name>.binary`), and `materialized-<provider>.json`/`graph.json` are local-environment state, not portable package content — none of it is meant to travel with the project's own source.
+
+`lineage enable` (`internal/cli.enableRef`) now calls `config.EnsureGitignored` the first time it creates a project's `.lineage/` — appending a `.lineage/` line to the project's `.gitignore`, creating the file if it doesn't exist. It's a no-op if some line already covers it (exact match, or a bare `.lineage`), and it only fires on that first creation, so a receiver who later removes the line on purpose isn't fought on every subsequent `enable`. There's no single `lineage init <project>` command today — `lineage enable` is the de facto point a project first gets a `.lineage/` at all, so that's where this lives; documented in `CONTRIBUTING.md` since there's no other natural home for the policy statement.
```
adkalhan: The comment says this handles a broader .gitignore rule, but the check only matches .lineage/ and .lineage exactly. Should we either handle the other common patterns that already ignore this directory, or narrow the comment to only the two patterns we actually support?

internal/cli/cli.go
```diff
@@ -721,6 +723,18 @@ func enableRef(ref, home string, autoApprove bool, stdin *bufio.Reader, stdout,
 		return false, err
 	}
 
+	// A project's .lineage/ mixes machine-local state (config.yaml's
+	// provider binary paths) with regenerable cache - it should never land
+	// in the receiver's repo. Only do this the first time a project gets a
+	// .lineage/ at all, so a receiver who deliberately removed the entry
+	// later isn't fought on every subsequent enable.
+	if isNewProject {
```
adkalhan: Do we want the gitignore protection to be one-time only? After the first enable, if someone removes .lineage/ from .gitignore, future enable calls won't restore it. I understand not wanting to fight an intentional removal, but this also means .lineage/ can become accidentally trackable later. Could we either document this behavior clearly or reconsider checking the ignore rule on each enable?

docs/decisions/0015-the-lineage-directory-is-a-versioned-enumerated-container.md
priyam-jain-2002: Addressed by narrowing the helper comment to the two patterns it actually recognizes: `.lineage/` and bare `.lineage`.

internal/cli/cli.go
priyam-jain-2002: Addressed by changing `enable` to reinforce the idempotent `.lineage/` ignore entry on every successful enable, plus updating the ADR/CONTRIBUTING wording and regression test to match.

adkalhan (approved):

PR #202, opened 2026-08-24
docs: reposition README around workflow packaging

priyam-jain-2002:
## Summary

- Rework the README opening around a compact Claude Code/Codex workflow-environment positioning statement.
- Move the author-to-receiver registry flow above architecture details so the first screen shows immediate proof.
- Preserve the package shape, command reference, safety principles, and existing guide/decision links while reducing repetitive framing.

## Linked Issue

Maintainer housekeeping

## Package Impact

- [x] Documentation only
- [ ] Tests only

## Safety

## Verification

Relevant test cases:

- README-only positioning change; no product behavior or CLI behavior is introduced.
- Full Go test suite run as a regression check.
- CLI help smoke check run to confirm README command surface still resolves.

```text
go test ./...
go run ./cmd/lineage --help
```

## Notes For Reviewers

This is a maintainer-authored documentation positioning change. The PR body was adjusted by the maintainer to satisfy the repository PR Policy parser; the content diff is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,43 @@
 # Lineage
 
-Lineage is an open-source local distribution layer for packaging and sharing AI
-agent workflows across Claude Code, Codex, and other coding agents.
+Package and share Claude Code and Codex workflow environments.
 
-It lets a prepared agent package travel as one environment: skills, workflows,
-agents, policies, references, setup material, and provider entrypoints can be
-bundled together, inspected, enabled, and launched from a receiver's machine.
-The package stays local and readable; the receiver keeps using their own agent
-tools.
+A reusable agent workflow usually depends on more than one prompt or skill:
+skills, workflow steps, agents, policies, references, setup files, declared
+capabilities, and provider entrypoints all shape how it behaves. Lineage keeps
+those pieces together as a local, inspectable package that another developer can
+add to their own project and run with their own agent tools.
 
-The goal is simple:
+```bash
+# author
+lineage package validate ./resume-workflow
+lineage login
+lineage package publish ./resume-workflow
 
-```text
-prepare once -> package safely -> share -> enable locally -> run with the user's own agent tools
+# receiver
+lineage add resume-workflow
+lineage run claude --dry-run
 ```
 
-Lineage does not try to replace the agent provider. It sits around local agent
-commands and makes the surrounding environment easier to package, review, and
-reproduce.
+`lineage add` fetches the package, shows what it contains, asks before enabling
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,56 +1,143 @@
...
-Lineage is an open-source local distribution layer for packaging and sharing AI
-agent workflows across Claude Code, Codex, and other coding agents.
+Package and share Claude Code and Codex workflow environments.
 
-It lets a prepared agent package travel as one environment: skills, workflows,
-agents, policies, references, setup material, and provider entrypoints can be
-bundled together, inspected, enabled, and launched from a receiver's machine.
-The package stays local and readable; the receiver keeps using their own agent
-tools.
+A reusable agent workflow usually depends on more than one prompt or skill:
```
adkalhan, line 5: It looks like some of the old README content wasn't removed when the new sections were introduced. For example, the "What Lineage Is For" bullets and the provider/safety sections contain duplicated content. Could we clean up the duplicated paragraphs so each concept is documented only once? This will make the README considerably easier to scan.

README.md
```diff
@@ -1,56 +1,143 @@
...
+```bash
+lineage login
+lineage package publish ./resume-workflow
+```
+
+The first publish of a package name claims it for your verified GitHub login.
+To ship an update, bump `version` in `lineage.yaml` and run
+`lineage package publish` again. Run `lineage whoami` to check the active
+identity, or `lineage logout` to clear it. For non-interactive use, set
+`LINEAGE_PUBLISH_TOKEN` to a GitHub-issued token with `read:user` access instead
```
adkalhan, line 92: Can we clarify why LINEAGE_PUBLISH_TOKEN only requires read:user access? Since this token is used for package publish, a reader may assume it needs write permissions. If read:user is intentionally sufficient because the registry only uses GitHub to authenticate/identify the publisher, it would be useful to state that explicitly here.

adkalhan (changes requested):

README.md
priyam-jain-2002: Addressed in the follow-up commit by trimming the duplicated safety/trust material from the README and pointing the later section at the canonical safety model instead.

README.md
priyam-jain-2002: Addressed by clarifying that `LINEAGE_PUBLISH_TOKEN` only identifies the publisher through GitHub, so repository write scopes are not required.

adkalhan (approved):

Issue #106, opened 2026-08-22
Compile provider-neutral package artifacts from the behavioral model


priyam-jain-2002:
## Problem statement

The first artifact generator should create the provider-neutral package structure. Provider-specific files can only be correct if the shared package manifest, workflows, skills, references, capabilities, and setup declarations are already coherent.

## Scope

- Generate `lineage.yaml` from behavioral-model metadata.
- Generate `workflows/<name>/WORKFLOW.md` with ordered steps compatible with #49.
- Generate `skills/<name>/SKILL.md` files for deterministic and agent-mediated steps.
- Generate references and setup metadata declared by the model.
- Preserve evidence comments or metadata where useful for review.

## Acceptance criteria

- Generated packages pass `lineage package validate` from #31.
- Workflow steps reference generated or required skills correctly.
- Package generation is deterministic for the same behavioral model.
- Tests cover a simple workflow, a multi-step workflow, references, and required external skills.
- The generator does not emit provider-specific instructions except through later provider emitters.

## Dependencies

- Depends on the behavioral model from the previous milestone.
- Builds on the workflow model used by #49.

## Non-goals

- No Claude-specific or Codex-specific context generation in this issue.
- No sanitization pass beyond producing structurally valid package files.
- No registry publish or pull changes.

## Sequencing

First child issue in artifact compilation. Provider-specific emitters should build on this package output.

priyam-jain-2002:
Maintainer follow-up from #275: when #106 introduces a serialized/persisted BehavioralModel input or output, define a canonical serialization rule for Decisions (for example, sorted by Decision.ID) and test byte-stable output. #275 intentionally leaves Decision order semantically unspecified because it does not serialize or persist models yet; do not rely on caller iteration order when adding that boundary.

Rushilkr77:
Starting on this. Findings from reading the merged #102–#104 work, and the implementation plan, for review before the PR.

**Plan (provider-neutral only, nothing Claude/Codex-specific)**
- `lineage analyze --model-out model.json` writes the full `BehavioralModel` (today `--yaml` only emits a lossy summary). `model.ParseModel` already reads it back.
[...]

Issue #104, opened 2026-08-22
Add Claude-assisted workflow analysis for package compilation


priyam-jain-2002:
## Problem statement

Human-authored agent workflows often encode behavior in prose and file layout. A deterministic scanner can inventory files, but it cannot reliably infer workflow intent, step ordering, or which instructions matter. Lineage needs an agent-assisted analysis stage that behaves like a constrained compiler, not a free-form rewrite.

## Scope

- Add a compilation analysis path that feeds the source inventory into Claude or a configured analysis provider.
- Constrain the provider to emit the behavioral model schema, including evidence links and unresolved decisions.
- Fail closed when the provider output is invalid, incomplete, or contradicts source evidence.
- Add dry-run/preview output so authors can inspect the inferred behavior before artifacts are generated.
- Keep provider access optional/configurable so deterministic tests can use fixtures.

## Acceptance criteria

- Analysis produces a valid behavioral model for a representative Claude/Codex source workspace.
- Invalid provider output is rejected with actionable diagnostics.
- Ambiguities are surfaced as decision items rather than silently filled in.
- Tests cover success, schema failure, missing evidence, and conflicting instruction cases using fixture provider responses.
- CLI output clearly separates source inventory, inferred behavior, and decisions needing author review.

## Dependencies

- Depends on source discovery and behavioral-model issues in this milestone.
- Should feed later artifact compilation milestones.

## Non-goals

- Do not allow the provider to edit the source workspace directly.
- Do not execute source scripts during analysis.
- Do not generate final package files in this issue.

## Sequencing

This is the final child issue in the workflow-understanding milestone. It unlocks artifact generation once the model is accepted.

priyam-jain-2002:
**Decision:** Confirming the proposed defaults: fail-closed on invalid/unparseable provider output rather than guessing, provider access stays optional and fixture-driven so tests are deterministic without live credentials, and no edits to the source workspace being analyzed.

Removing `needs:decision`. Blocked on #103's schema landing first, so holding off on `help wanted` until then.

Rushilkr77:
Opened #277, built on #103's BehavioralModel (#275) and #102's inventory.

Confirms the three defaults from the pinned decision comment:

- Fail-closed: `model.ParseModel` + `model.Validate` reject malformed, invalid, or evidence-contradicting provider output. No new validation logic, just reuses #103's existing collect-everything validator.
[...]

Issue #103, opened 2026-08-22
Define the behavioral workflow model for compiled packages


priyam-jain-2002:
## Problem statement

The compiler needs a provider-neutral intermediate representation. Without a clear model, artifact generation will either copy source files blindly or bake provider-specific assumptions into the analysis step.

## Scope

- Define a versioned behavioral model for compiled workflows.
- Represent workflow intent, ordered steps, step inputs/outputs, required skills, required scripts/tools, references, setup requirements, validation gates, and unresolved decisions.
- Represent which source artifacts justify each part of the model.
- Define how this model maps onto existing package concepts: manifest exports, workflows, skills, agents, policies, references, capabilities, and setup material.
- Add schema validation and fixtures.

## Acceptance criteria

- The model is versioned and documented.
- A sample messy workflow can be represented without provider-specific fields.
- Every step can reference source evidence from the discovery inventory.
- Missing or ambiguous behavior is expressible as an explicit decision item.
- Tests reject malformed models and accept representative valid models.

## Dependencies

- Depends on the source inventory issue in this milestone.
- Builds on #49, which established the current ordered workflow step concept.
- Should remain compatible with #31 package validation.

## Non-goals

- No Claude/Codex artifact emission.
- No receiver setup execution.
- No behavioral test runner yet.

## Sequencing

Land after source discovery and before the agent-assisted analysis path, because the analysis prompt and parser need a target schema.

priyam-jain-2002:
**Decision:** Confirming this issue's own proposal — a versioned, schema-validated behavioral model with a `schema` field, mirroring `lineage.yaml`'s existing convention (ADR 0005), with evidence links back to #102's inventory output. Ambiguity in a given workflow should surface as an explicit decision item in the model, not a silent guess.

Removing `needs:decision`; adding `help wanted` — this blocks #104 and everything downstream, so it's a good next pickup once #102 lands, but the schema design means it's not entry-level.

Rushilkr77:
Opened #275 implementing this: internal/model.BehavioralModel, the versioned provider-neutral schema this issue asked for, built on top of #203's evidence inventory. The PR description covers the design; ADR 0017 records the schema decision.

Issue #101, opened 2026-08-22
Goal: Compile existing agent workflows into portable Lineage packages


priyam-jain-2002:
## Problem statement

`lineage package` currently assumes the source already looks like a clean Lineage package. Real agent workflows are usually spread across `CLAUDE.md`, `AGENTS.md`, `SKILL.md`, scripts, prompts, references, and undocumented sequencing. Copying those files preserves structure, but not necessarily behavior.

## Goal

Turn packaging into a behavior-preserving compilation flow:

1. Discover source workflow evidence.
2. Ask an agent/compiler to reconstruct the intended behavior.
3. Persist a structured behavioral model.
4. Hand that model to later artifact, sanitization, and validation stages.

## Scope

- Add a compilation path for existing Claude/Codex-style workspaces.
- Define what source artifacts are inspected and how their influence is recorded.
- Produce an intermediate behavioral model before generating package files.
- Keep the model provider-neutral so later milestones can emit Claude Code, Codex, and future provider artifacts.
- Reuse the current package manifest and validation foundation from #31 instead of creating a parallel package format.

## Acceptance criteria

- A messy source workspace can be analyzed without mutating it.
- The compilation output includes workflow intent, ordered steps, required inputs, scripts/tools, references, provider assumptions, and validation hooks.
- Ambiguous or conflicting behavior is reported as a decision item instead of silently guessed.
- Tests cover discovery from `CLAUDE.md`, `AGENTS.md`, `SKILL.md`, scripts, prompts, and references.
- Documentation explains the compilation pipeline and where it fits before artifact generation.

## Dependencies

- Builds on `lineage package validate` from #31.
- Builds on the minimal workflow execution model from #49.
- Should inform, but not block, provider materialization work already completed in #18 and #19.

## Non-goals

- Do not execute arbitrary source scripts during analysis.
- Do not require a provider-specific artifact layout at this stage.
- Do not promise perfect semantic recovery for every arbitrary repository.

## Sequencing

This milestone should land before cross-agent artifact compilation, sanitization, and behavioral validation. The behavioral model is the contract those later stages consume.

priyam-jain-2002:
Child issue sequence:

1. #102 - Add source workspace discovery and evidence inventory for package compilation
2. #103 - Define the behavioral workflow model for compiled packages
3. #104 - Add Claude-assisted workflow analysis for package compilation

Foundation links: #31, #49, #18, #19.

priyam-jain-2002:
[...]

PR #203, opened 2026-08-24
Add source workspace discovery and evidence inventory

### LINKED ISSUE #102: Add source workspace discovery and evidence inventory for package compilation
## Problem statement

Before Lineage can compile a workflow, it needs a deterministic inventory of the source workspace. Today package discovery starts from `lineage.yaml`; this new path must inspect existing agent-workflow folders that may not already be valid Lineage packages.

## Scope

- Add a read-only discovery pass for source workspaces.
- Inventory agent instruction files such as `CLAUDE.md`, `AGENTS.md`, and nested `SKILL.md`/`WORKFLOW.md` files.
- Inventory scripts, prompts, reference files, setup notes, and existing package manifests when present.
- Classify each artifact as instruction, executable helper, reference, setup material, package metadata, or unknown.
- Record file paths, content digests, size, language/type, and why each artifact was included.

## Acceptance criteria

- Discovery runs without creating, editing, deleting, or executing source files.
- Output is deterministic for the same source tree.
- Ignored directories such as `.git`, dependency folders, generated archives, and common build output are skipped by default.
- Tests cover mixed Claude/Codex workspaces, nested skills, scripts, references, and ignored paths.
- The inventory can be serialized for the later behavioral-model issue.

## Dependencies

- Should reuse existing package safe-path assumptions from #31 and related validation work.

## Non-goals

- No agent/LLM analysis in this issue.
- No package artifact generation in this issue.
- No execution of discovered scripts.

## Sequencing

First child issue in the workflow-understanding milestone. The behavioral model and Claude-assisted analysis issues consume this inventory.

Rushilkr77:
## Summary

Adds `internal/inventory`, a read-only pass that walks a source workspace that isn't a Lineage package yet (no `lineage.yaml` required) and produces a deterministic inventory: every file's kind, why it was classified that way, and which markdown files literally name it in their text.

Closes #102.

This is the first link in the chain #106 needs: #102 (this PR) -> #103 (behavioral model) -> #104 (agent-assisted analysis) -> #106 (artifact compilation). No provider adapters, materialization, or CLI surface touched. Pure new package, additive only.

## Design

Two passes, run in `Discover(root)`:

**Classify.** Each file gets a `Kind` (`instruction` / `executable_helper` / `reference` / `setup_material` / `package_metadata` / `unknown`) by name/path heuristic, not content sniffing or an LLM. A file matching no rule is honestly `unknown` with a stated reason, not a forced guess.

**Cross-reference.** Every markdown file is scanned line by line for literal path/filename mentions of other inventoried files. Each hit becomes a `Citation{FromPath, Line, Snippet}`, recorded on both sides (`Entry.Mentions` outgoing, `Entry.ReferencedBy` incoming) as arrays, since one script is often cited by several workflows or skills. Citing is gated on being markdown, not on `Kind`: a file classify() can't confidently label (`AGENT.md`, a `README.md` inside a skill folder) still carries real citation evidence and must not be silently excluded.

This is plain string matching, not understanding. `scripts/deploy.sh` written out gets caught; "run the deploy script" with no filename doesn't. That gap is intentional: #102's non-goals rule out LLM analysis, and resolving paraphrased references is #104's job. The citation structure exists so #103/#104 can point at exact evidence, a line and a capped snippet, instead of re-parsing whole files into an LLM prompt.

## Decisions made

- **Bare-filename matching only fires when the basename is unique across the tree.** Found by testing against a real repo (`newsflow`): a debugging doc naming `pipeline.log` and `curated_articles.json` matched every dated output file sharing those names, turning one real citation into hundreds of false ones. Ambiguous basenames still match via full relative path.
- **Directory-based instruction detection is deliberately shallow and provider-prefix-aware.** `skills/<name>/*.md`, `workflows/*.md` or `workflows/<name>/*.md`, and `agents/*.md` (optionally under `.claude/` or `.agents/`, mirroring `internal/provider`'s registry) count as that unit's own instructions even with non-canonical filenames. Anything nested deeper doesn't: a first version that matched any depth caught real false positives (eval-log `outputs/response.md`, `templates/*.md`) in a cloned public repo.
- **Default ignore list covers `.git`, `node_modules`, `vendor`, common build/cache dirs, and `venv`/`.venv`.** `venv` specifically came from testing against a real Python project whose `bin/` is full of symlinks; without the skip, one uncommon directory aborted the whole run.
- **Symlinks are refused outright**, matching the existing pattern in `internal/packages.contentFiles` and `internal/materialize.copyDir`: never followed, never silently skipped.

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Safety

## Verification

Relevant test cases:

- `TestDiscoverDeterministic` - same tree, two runs, identical result
- `TestDiscoverSkipsIgnoredDirs` - `.git`/`node_modules` never appear in output
- `TestDiscoverClassifiesRealisticShapes` - per-kind classification across a mixed fixture
- `TestDiscoverCitationSingleReference` / `TestDiscoverCitationFanIn` - exact line/snippet citation, and one file cited by two different sources
- `TestDiscoverOrphanScriptUnreferenced` / `TestDiscoverUnknownIsHonestNotGuessed` - unreferenced and unclassifiable stay honest, not guessed
- `TestDiscoverAmbiguousBasenameOnlyMatchesFullPath` - regression for the `newsflow` bare-name fan-out bug
- `TestDiscoverRecognizesProviderPrefixedAgentFiles` - regression for the `.claude/agents/*.md` gap found in a cloned public repo
- `TestDiscoverRejectsSymlink` - symlinked content errors rather than following
- `TestMessinessLevel1` through `TestMessinessLevel5` - the same real workflow (`examples/resume-workflow`) rewritten at five levels of ambiguity, from a single file with inlined scripts down to proto-skill folders, each asserting the specific pattern that level exercises

```text
go build ./...
go vet ./...
go test ./...
```

Also manually verified against three real, unmodified repos (not part of the automated suite): this repo's own `examples/resume-workflow`, a cloned Python project with a real `venv/`, and two cloned public Claude Code repos (`ChrisWiles/claude-code-showcase`, `meetrais/claude-agent-skills`); each surfaced a real bug now fixed and covered above.

- N/A, see test cases above.

## Notes For Reviewers

`internal/inventory` is a new package with no CLI entrypoint by design: it's an internal API consumed by #103/#104 once they exist. #102 is only buildable up to this contract; nothing downstream can be wired up until #103's behavioral model schema lands.


Rushilkr77:
How this sets up the rest of the chain (#103 → #104 → #106):

**#103 (behavioral model schema)** consumes `Inventory` directly as its evidence layer. Every field exists because #103 needs it: `Kind` gives it a pre-sorted candidate list instead of raw files: instructions, executable helpers, references, setup material, and an explicit `unknown` punch list for whatever couldn't be classified deterministically. `Digest` lets the model (and everything after it) prove it analyzed exactly this content, and skip re-analysis when nothing changed. `Mentions`/`ReferencedBy` give #103 a real, if incomplete, dependency graph between files, already resolved, so building the model's ordered steps doesn't start from zero.

**#104 (agent-assisted analysis)** is where the boundary this PR draws on purpose gets closed. Everything #102 could establish mechanically, it did: literal path citations, deterministic classification, honest `unknown` flags. Everything it couldn't, an LLM is allowed to (#104's non-goal in #102 is #104's actual job): resolving "run the deploy script" with no filename, deciding what an `unknown` file actually is, connecting a workflow step to a skill by name instead of by literal path. #104's prompt can be built almost entirely from `Inventory` plus targeted file reads at the exact `Citation` line/snippet, instead of dumping full file contents at a model and hoping it finds the relevant part.

**#106 (compile provider-neutral package artifacts)** is the far end. It never touches `internal/inventory` directly; it consumes whatever #103's model produces. But #106 only exists because #102 → #103 → #104 gets a messy workspace into a real behavioral model first, so #106 has something structured to compile instead of copying source files blindly (which is the exact failure mode #101, the milestone-11 parent goal, calls out as insufficient).

Net effect: this PR is the only piece of the chain that touches real, arbitrary, untrusted source trees. Everything downstream operates on the structured `Inventory` this produces, not on raw file walks of their own.

priyam-jain-2002 (changes requested):
Thanks for the detailed implementation and for keeping the #102 boundary clear. The read-only walk, deterministic ordering, ignored directories, symlink refusal, and fixture coverage are all pointing in the right direction.

I think this needs one API-contract fix before merge: `Entry.Mentions` records `Citation{FromPath, Line, Snippet}` on the citing entry, but the citation does not include the target path. That means consumers looking at an instruction file cannot tell which inventoried artifact each outgoing mention points to. It also makes multiple targets found on the same line indistinguishable, because they would produce identical citations in `Mentions`.

Since this package is intended to be the evidence layer for #103/#104, the graph should be round-trippable from either side. Please include the target path in the citation model, for example `ToPath`/`TargetPath`, populate it in `crossReference`, and add a test that asserts `Mentions` identifies the cited artifact, including a case where one line names more than one file.

I also fixed the mechanical PR Policy issue in the PR body by adding `Closes #102`. Local verification passed with `go test ./internal/inventory ./...`.

priyam-jain-2002 (changes requested):
Thanks again for the thoughtful scope control here. I did a deeper review pass across functionality, security/path behavior, performance shape, tests, and downstream API usability. The package is nicely contained, CI is green, and the read-only/non-LLM boundary is the right direction for #102.

I still think this should wait on a few correctness fixes before merge:

1. `Entry.Mentions` needs to identify the mentioned target. Today `Citation` only has `FromPath`, `Line`, and `Snippet`, so outgoing mentions on the citing file cannot tell consumers which artifact was mentioned. This also collapses distinct targets on the same line into indistinguishable citations. Since #103/#104 are supposed to consume this as evidence, please add a target field such as `ToPath`/`TargetPath`, populate it in `crossReference`, and add tests for outgoing mentions, including one line that names two different files.

2. Citation matching currently uses raw `strings.Contains` for both full relative paths and basenames. That produces false evidence for substring matches: a line mentioning `scripts/deploy.sh.bak` will also cite `scripts/deploy.sh`. Because these citations are meant to be evidence, they should match path-like/filename token boundaries rather than arbitrary substrings. Please add a regression test for suffix/prefix false positives and tighten the matcher accordingly.

3. A symlink passed as the discovery root is not refused; it returns a successful empty inventory because the `path == root` case returns before the symlink check. The PR description says symlinks are refused outright, and callers can reasonably pass a source workspace path that happens to be a symlink. Please either reject a symlink root explicitly before `WalkDir`, or deliberately resolve it and document that behavior, with a test.

Non-blocking consideration: the cross-reference pass is O(markdown lines * files) and reads each markdown file fully into memory. That may be acceptable for v1, but it would be good to leave a comment or follow-up if this is expected to run against large arbitrary workspaces.

Verification I ran locally:

```text
go test ./internal/inventory ./...
go vet ./...
```

`go test` and `go vet` both pass on the current PR head. The requested changes are about evidence correctness and edge-case behavior rather than current CI failures.

priyam-jain-2002 (changes requested):
Thanks @Rushilkr77. I reviewed the latest commits carefully against #102 and the previous review.

The earlier blockers are substantially addressed: `Citation` is now a self-describing edge with `ToPath`, `Column`, `MatchedText`, and `MatchKind`; basename matches are no longer raw substring matches; symlink roots are refused before walking; and the tests now cover those regressions directly. The schema stamp and snippet-copy cleanup are also good additions for the downstream #103/#104 handoff.

There is one remaining correctness issue before this should merge:

- The path-token matcher still accepts a match when the candidate path is only the suffix of a different path segment. Because `/` is deliberately not treated as a continuation byte, `references/data.csv` is cited when the markdown text says `archive/references/data.csv`. I reproduced this locally with a temporary regression test: a workspace containing both `references/data.csv` and `archive/references/data.csv`, with `CLAUDE.md` saying only `archive/references/data.csv`, still adds `ReferencedBy` to `references/data.csv` with `MatchedText: archive/references/data.csv`. For an evidence inventory, that is false evidence rather than just a missed heuristic.

Please tighten this boundary so path-prefix tolerance allows real relative prefixes like `../` / `../../` without allowing arbitrary directory prefixes to cite a shorter suffix path. A regression test next to `TestDiscoverCitationRequiresTokenBoundary` / `TestDiscoverCitationSurvivesAdjacentPunctuation` would be enough: the shorter `references/data.csv` target should not be cited when only `archive/references/data.csv` is written, while the existing `../../references/data.csv` case should keep passing.

Also please update the branch with current `develop` before the next review pass. The merge is clean locally, and after merging `develop` locally I verified:

```text
go test ./...
go vet ./...
go build ./...
git diff --check origin/develop...HEAD
gofmt -l internal/inventory/*.go
```

Those passed in the merged local checkout. GitHub currently reports no checks on the PR head, so getting the branch current should restore the normal CI signal.

priyam-jain-2002 (changes requested):
Thanks for the latest round. I re-reviewed the current head (`46d5b3b`) against #102 and the previous review comments. The earlier blockers are addressed: citations are self-describing, substring/path-boundary false positives are covered, symlink roots are refused, the branch is current with `develop`, and GitHub checks are green.

I found one remaining evidence-correctness issue before merge:

- Relative path references with `.` / `..` are bounded against the workspace root but are not actually resolved from the citing markdown file’s directory. `classifyReference` strips leading `.` and `..` segments, then compares the remaining suffix against every target path. That means a nested instruction can cite the wrong root-level artifact. Repro: create `skills/foo/SKILL.md` containing `Run ../scripts/deploy.sh.` and a root file `scripts/deploy.sh`. From `skills/foo`, that written path resolves to `skills/scripts/deploy.sh`, not root `scripts/deploy.sh`, but `Discover` records a `ReferencedBy` edge on root `scripts/deploy.sh` with `MatchedText: ../scripts/deploy.sh` and `MatchKind: path`.

Because `ReferencedBy` / `Mentions` are intended to be exact evidence for #103/#104, this should not manufacture an edge for a path that resolves elsewhere. Please resolve navigated path tokens relative to `path.Dir(FromPath)` before comparing to candidate paths, or otherwise reject navigated references that cannot be proven to resolve to the target. A regression test should cover the false-positive case above, while keeping the intended `../../references/data.csv` from `skills/foo/SKILL.md` behavior passing.

Local verification on the PR head:

```text
go test ./internal/inventory
go test ./...
go vet ./...
gofmt -l internal/inventory/*.go
git diff --check origin/develop...HEAD
```

All of those pass; the requested change is about citation correctness rather than CI failure.

Rushilkr77:
Fixed in `64404fd`, plus four more from an audit of the same class of problem.

**Reported bug.** `classifyReference` bounded how far `..` could climb (`ups > citerDepth` rejects) but never computed where it landed, so navigated references fell through to the fuzzy suffix matcher. `../scripts/deploy.sh` in `skills/foo/SKILL.md` has `ups=1`, inside the bound, so it passed the guard and suffix-matched root `scripts/deploy.sh`. It resolves to `skills/scripts/deploy.sh`.

Any reference with a leading `.` or `..` now resolves against the citer's directory and must equal `target.path` exactly, no `MatchBasename` fallback. This also covers a bare `./run.sh`, which you didn't ask about.

The old path-shape sweep missed this because every navigated row tested `ups == citerDepth` (lands at root) or `ups > citerDepth` (rejected). A partial unwind is the only shape where wrong ancestor and true target differ. Both directions are table cases now, and your `../../references/data.csv` case still passes.

Against this repo's own docs the resolved target set is byte-for-byte unchanged; several citations upgrade from `MatchBasename` to `MatchPath`. `../architecture.md` in `docs/guides/lineage-package.md` was previously correct only because that basename is unique in the tree.

**Four from the audit:**

1. `MatchBasename` accepted extensionless basenames, so a file named `run` got cited by "Please run the tests". Guard checks written segment count. Anything with a `/` is unaffected.
2. `Snippet` took a line's first 200 bytes unconditionally, so on a long line with the match past the cutoff it showed unrelated text while `Column`/`AsWritten` pointed correctly. Now centers the window on the match.
3. `MatchedText` renamed to `AsWritten`, with a non-goals section in the package doc. `ToPath` was always authoritative, but the old name invites treating it as resolvable when for a basename edge it is deliberately partial. `Column` is now documented as a byte offset. Zero importers, so the rename is free.
4. Package doc states the markdown-only boundary: a file referenced solely via a manifest's `entrypoints` field produces no citation, so empty `ReferencedBy` does not mean orphaned.

Each code fix was verified by disabling it and confirming its regression test fails. Branch is current with `develop`.

```text
go build ./...
go vet ./...
gofmt -l internal/inventory/*.go
go test ./...
```

Ready for re-review.

priyam-jain-2002 (approved):
Approved. I re-reviewed the latest head against #102 and the prior review rounds.

The earlier blockers are addressed: citations are self-describing, substring/path-boundary false positives are covered, symlink roots and unusable roots fail loudly, relative `.`/`..` references resolve from the citing markdown file's directory, ambiguous basenames are surfaced, and snippets are centered on the matched reference.

Local verification passed on the reviewed head:

```text
go test ./internal/inventory
go test ./...
go vet ./...
gofmt -l internal/inventory/*.go tests/receiver_flow_test.go
git diff --check origin/develop...HEAD
go test -coverprofile=coverage.out ./...
```

Only remaining note is non-blocking documentation polish: the `Citation.Snippet` field comment says it always contains `AsWritten`, while the helper's fuller comment correctly qualifies that guarantee for realistic matches shorter than `maxSnippetLen`. Behavior is sound because `ToPath`, `AsWritten`, `Line`, and `Column` remain authoritative. Good to merge.

PR #205, opened 2026-08-24
packages: preserve the executable bit through export and import

### LINKED ISSUE #165: Export/import discards the executable bit on package content
## Problem

`internal/packages/export.go:75` hardcodes `Mode: 0o644` for every tar entry, and `internal/packages/import.go:130` (`writeArchiveFile`) opens with the same fixed mode. Any package content meant to carry an executable bit (e.g. a helper script under `adapters/`) silently loses `+x` on export and stays lost on import.

## Fix direction

Preserve the source file's mode bits (masked to a safe range) through export and import instead of hardcoding `0o644`.

## Confidence

High.

MsfPablo:
## Summary

`addTarFile` hardcoded `Mode: 0o644` on every tar entry, and `writeArchiveFile` opened every extracted file at `0o644`, so a helper script under `adapters/` lost `+x` on export and never got it back on import. `copyTree` — the staging copy import runs afterwards — hardcoded it a third time, so fixing only the first two would still have dropped it.

All three now route through one `ContentMode` helper.

## The mode is collapsed, not copied

`ContentMode` maps a source mode to exactly one of two values: `0o755` if any execute bit is set, `0o644` otherwise. It deliberately does not carry the source mode through, for two reasons:

- **Package content is untrusted.** Copying `hdr.Mode` would let a crafted archive ask for setuid, setgid, sticky, or group/world-writable bits on the receiver's disk. Only the exec bit survives, and only as the fixed `0o755`.
- **Archives must stay deterministic.** `deterministicTime` exists so identical content exports byte-identical; copying a contributor's umask-dependent mode would reintroduce exactly the variance that comment is guarding against.

## Linked Issue

Closes #165

## Package Impact

## Safety

Import's permission surface narrows here rather than widening: previously `hdr.Mode` was ignored entirely, and now it is read but clamped, so the worst an archive can request is `0o755`.

## Verification

Relevant test cases:

- `TestExportImportPreservesExecutableBit` (new) — a `0o755` script and a `0o644` doc through a full `Export`/`Import` round trip; both modes land correctly. Fails on `develop` with `adapters/run.sh mode = 644 after a round trip, want 755`. Skipped on Windows.
- `TestContentModeCollapsesUnsafeBits` (new) — table covering `0o600`, `0o700`, `0o777`, `0o666`, and setuid/setgid inputs, asserting nothing outside the two allowed values comes back.
- `TestExportIsDeterministic` — unchanged and still passing, which is the check that matters most for the collapse decision above.

```text
go test ./...
(all packages ok)
```

## Notes For Reviewers

`ContentMode` is exported because import and export live in the same package but the concept is worth naming for anyone adding a third write path later — `copyTree` was exactly that case and was already out of sync. Happy to unexport it if you'd rather keep the surface minimal.

`addTarFile` now does an `os.Stat` it did not do before. It already read the whole file into memory, so this is not a meaningful cost.

### DIFF
--- a/internal/packages/export.go
+++ b/internal/packages/export.go
@@ -63,17 +63,39 @@ func Export(dir string, w io.Writer) error {
 	return nil
 }
 
+// ContentMode collapses a file's permissions to one of exactly two values:
+// 0o755 when any execute bit is set, 0o644 otherwise. Export, import, and
+// the staging copy all route through it so a helper script keeps its
+// executable bit across a round trip.
+//
+// It collapses rather than copies the source mode for two reasons. Package
+// content is untrusted, so an archive must never be able to ask for setuid,
+// setgid, sticky, or group/world-writable bits on a receiver's disk — only
+// the exec bit survives, and only as the fixed 0o755. And an archive has to
+// stay byte-identical for identical content (see deterministicTime), which
+// copying a contributor's umask-dependent mode would break.
+func ContentMode(m os.FileMode) os.FileMode {
+	if m.Perm()&0o111 != 0 {
+		return 0o755
+	}
+	return 0o644
+}
+
 func addTarFile(tw *tar.Writer, absPath, name string) error {
 	data, err := os.ReadFile(absPath)
 	if err != nil {
 		return fmt.Errorf("read %s: %w", name, err)
+	}
+	info, err := os.Stat(absPath)
+	if err != nil {
+		return fmt.Errorf("stat %s: %w", name, err)
 	}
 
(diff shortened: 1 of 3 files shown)


priyam-jain-2002 (approved):
Approved. Preserving only the executable bit through a fixed ContentMode keeps helper scripts runnable while still collapsing unsafe archive permissions to a deterministic, constrained mode. The export/import round-trip and unsafe-bit collapse tests cover the risk well. PR Policy and Test are green after automatic branch update, and I verified go test ./internal/packages plus go test ./... locally.

PR #206, opened 2026-08-24
packages: reject an explicit schema: 0 instead of coercing it

### LINKED ISSUE #166: schema: 0 in a manifest is indistinguishable from an absent schema field
## Problem

`internal/packages/manifest.go:143-147`:

\`\`\`go
// schema 0 means the manifest predates the schema field; treat that as
// schema 1, the only schema that ever existed before it was added.
if manifest.Schema == 0 {
    manifest.Schema = CurrentSchema
}
\`\`\`

YAML-unmarshaling an absent int field and an explicit `schema: 0` both leave `Schema == 0`. A manifest that explicitly (or accidentally) declares `schema: 0` is silently coerced to the current schema rather than rejected, even though `0` never existed as a real schema value.

## Fix direction

Use a pointer (`*int`) or a sentinel to distinguish "field absent" from "field explicitly zero," or explicitly document/accept that `0` is reserved to mean "absent" and reject any other invalid value clearly.

## Confidence

High on the mechanism, low on current impact — worth fixing before `0` could be reused for something else.

MsfPablo:
## Summary

YAML-unmarshaling an absent `schema` field and an explicit `schema: 0` both leave `manifest.Schema == 0`, so the int alone could not tell them apart. A manifest declaring `schema: 0` — a value that never existed as a real schema — was silently coerced to the current schema and accepted, while every other invalid value (`schema: 2`) was rejected.

`LoadManifest` now decodes the field a second time as `*int` purely to answer "was this key present?". Absent still means the manifest predates the schema field and is treated as schema 1; an explicit `0` now falls through to the existing "declares schema %d, but this build only understands schema %d" error.

## Linked Issue

Closes #166

## Package Impact

## Safety

A manifest is untrusted input, and this makes one previously-accepted malformed value fail closed. Nothing that was valid before becomes invalid.

## Verification

Relevant test cases:

- absent schema -> accepted, normalized to `CurrentSchema` (the pre-schema-manifest path, unchanged)
- `schema: 1` -> accepted
- `schema: 0` -> rejected; fails on `develop` with `error = nil, want a rejection`
- `schema: 2` -> still rejected, confirming the existing path did not regress

```text
go test ./...
(all packages ok)
```

## Notes For Reviewers

The issue offered a pointer field or a sentinel. I used a pointer, but in a local probe struct rather than on `Manifest` itself — changing `Manifest.Schema` to `*int` would ripple into `SaveManifest`, `DefaultManifest`, and every reader of the field, for a question only `LoadManifest` ever asks. This keeps the public type and the on-disk output identical.

Cost is a second `yaml.Unmarshal` over an already-in-memory manifest, which is negligible at manifest size. If you would rather have the pointer on the struct as a deliberate API statement, say so and I will switch it.

The error message for the zero case is the existing schema-mismatch one, which reads correctly ("declares schema 0, but this build only understands schema 1"). I did not add a special-cased message for `0`, but it is a one-liner if you want the reserved-value meaning spelled out.

### DIFF
--- a/internal/packages/manifest.go
+++ b/internal/packages/manifest.go
@@ -163,9 +163,20 @@ func LoadManifest(dir string) (Manifest, error) {
 	if err := validateIdentifier("version", path, manifest.Version); err != nil {
 		return Manifest{}, err
 	}
-	// schema 0 means the manifest predates the schema field; treat that as
-	// schema 1, the only schema that ever existed before it was added.
-	if manifest.Schema == 0 {
+	// An absent schema field and an explicit `schema: 0` both leave
+	// manifest.Schema at 0, so the int alone cannot tell them apart. Decode
+	// the field again as a pointer to distinguish them: an absent field
+	// means the manifest predates the schema field and is treated as schema
+	// 1, the only schema that ever existed before it was added, while an
+	// explicit 0 is a value that never existed and is rejected rather than
+	// silently coerced.
+	var probe struct {
+		Schema *int `yaml:"schema"`
+	}
+	if err := yaml.Unmarshal(data, &probe); err != nil {
+		return Manifest{}, fmt.Errorf("parse manifest %s: %w", path, err)
+	}
+	if probe.Schema == nil {
 		manifest.Schema = CurrentSchema
 	}
 	if manifest.Schema != CurrentSchema {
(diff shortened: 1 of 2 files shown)


priyam-jain-2002 (approved):
Approved. The local probe cleanly distinguishes an absent schema from an explicit `schema: 0` without changing the public Manifest shape, and the table test covers legacy, current, zero, and future schema cases. PR Policy and Test are green after the mechanical body fix and automatic branch update; I also verified go test ./internal/packages plus go test ./... locally.

PR #208, opened 2026-08-24
tests: add end-to-end receiver-flow integration test

### LINKED ISSUE #207: Add end-to-end receiver-flow integration test
## Problem

`tests/README.md` reserves this directory for "higher-level fixtures and integration tests as package export/import behavior matures," but nothing lives there yet — coverage of the full distribution story (author creates + validates a package → exports → a receiver on a separate `LINEAGE_HOME`/project imports → inspects → enables → runs the whole package → runs a single workflow with scoped materialization → a plain run restores the full set → lists → disables → doctors clean) is currently spread across many package-level unit tests, none of which exercise the real CLI entrypoint (`cli.Execute`) start to finish the way an actual receiver would.

This gap was originally tracked under #12 (closed 2026-08-23 as part of the general CI-integration-coverage push), but this specific end-to-end fixture never landed — it was written on a local branch (`test/receiver-flow`, Aug 19) built by locally combining two other unmerged branches just to have everything the test needed available at once, with a note that it should move to a clean branch off `develop` once those merged. They have since merged, but the branch was never rescued.

## Ask

Add a `tests/receiver_flow_test.go` fixture covering the full round trip above through the real CLI surface, so a regression in how the commands compose (not just each command in isolation) gets caught in CI.

## Confidence

High — the test already exists in a working form and just needs the module path and a since-changed `disable` confirmation gate (#187 added a `--yes` requirement) updated to match current `develop`.

priyam-jain-2002:
## Summary

Adds `tests/receiver_flow_test.go`, a single fixture that recreates the full distribution story through the real CLI entrypoint (`cli.Execute`) rather than testing each command's internals in isolation: an author creates + validates a two-skill package with an ordered workflow, exports it, a receiver (separate `LINEAGE_HOME` and project, sharing nothing but the archive) imports, inspects, enables, runs the full package, runs just one workflow (scoped materialization), confirms a plain run afterward restores the full set, then lists, disables, and doctors clean.

This was originally written on a local branch on 2026-08-19 that combined two other branches purely so everything it needed was available to write against at once, with a note to move it to a clean branch off `develop` once those merged — they have, but it was never rescued. Ported here with two updates for drift since then:

- Module path is `github.com/agentic-lineage/lineage` now (was `github.com/lineage-dev/lineage`).
- `disable` gained a confirmation gate in #187; the test now passes `--yes`, consistent with how it already calls `run`/`workflow run`.

## Linked Issue

Closes #207

- [x] This PR targets `develop`.

## Package Impact

- [x] Tests only

## Safety

## Verification

Relevant test cases:

- `TestReceiverFlow`

```text
go build ./... && go vet ./... && go test ./...
```

- N/A, this PR is the test.

## Notes For Reviewers

Everything downstream of the module-path rename and the `disable --yes` gate applied cleanly with no other adaptation needed — the rest of the CLI surface this test exercises hasn't drifted from what the 08-19 version assumed.


tests/receiver_flow_test.go
```diff
@@ -0,0 +1,252 @@
...
+	inspectOut := run(t, receiverHome, receiverProject, nil, "inspect", "field-notes")
+	if !strings.Contains(inspectOut, "field-notes@0.1.0") || !strings.Contains(inspectOut, "workflows: field-report") {
+		t.Fatalf("inspect output = %q, want package identity and its workflow", inspectOut)
+	}
+	if _, err := os.Stat(config.ProjectConfigPath(receiverProject)); !os.IsNotExist(err) {
+		t.Fatal("inspect must not have enabled anything yet")
+	}
+
+	// Confirm identity survived the trip intact before doing anything
+	// else: the receiver's copy must be byte-for-byte what the author
```
adkalhan: The digest comparison below is a good check, but this comment says we're verifying the package is "byte-for-byte" identical. We're comparing the discovered package digests rather than the archive/file bytes directly, so could we reword this to say the imported package has the same content/identity as the author's package?

tests/receiver_flow_test.go
```diff
@@ -0,0 +1,252 @@
...
+	// ordered workflow, and validate it before it ever leaves this
+	// machine. ---
+	authorHome := filepath.Join(root, "author-home")
+	pkgDir := filepath.Join(root, "field-notes")
+
+	run(t, authorHome, root, nil, "package", "init", pkgDir)
+
+	writeFile(t, filepath.Join(pkgDir, "skills", "collect", "SKILL.md"), "---\nname: collect\ndescription: Collect field notes from the receiver.\n---\n\nAsk the receiver for the essentials before doing anything else.\n")
+	writeFile(t, filepath.Join(pkgDir, "skills", "summarize", "SKILL.md"), "---\nname: summarize\ndescription: Summarize collected field notes.\n---\n\nTurn the collected notes into a short, honest summary.\n")
+	writeFile(t, filepath.Join(pkgDir, "workflows", "field-report", "WORKFLOW.md"), "---\nsteps:\n  - collect\n  - summarize\n---\n\n# Field Report\n\n1. Collect notes.\n2. Summarize them.\n")
```
adkalhan, line 45: Could we add a third skill to the package that is not part of field-report? Right now the package contains only collect and summarize, and the workflow contains both of them, so the workflow run produces the exact same set of skills as the full package run. This means the test would still pass if workflow scoping were completely ignored. For example, adding an archive skill to the package but leaving it out of this workflow would let us assert that archive is absent after workflow run and restored after the subsequent full run

adkalhan (changes requested):

tests/receiver_flow_test.go
priyam-jain-2002: Reworded this in b2cad11 to describe matching package content identity via discovered package digests, rather than claiming byte-for-byte archive/file equality.

tests/receiver_flow_test.go
priyam-jain-2002: Addressed in b2cad11. I added an archive skill that is part of the package but not part of the field-report workflow. The test now asserts archive is materialized during the full package run, absent after the scoped workflow run, restored by the next full run, and cleaned up on disable.

adkalhan (approved):

PR #209, opened 2026-08-24
ci: run Go tests on Windows

c436zhan:
## Summary

Adds a small OS matrix to the Go test workflow so pull requests and protected-branch pushes run `go test -coverprofile=coverage.out ./...` on both `ubuntu-latest` and `windows-latest`.

The coverage summary step now explicitly uses Bash across the matrix, and coverage artifacts are named per OS to avoid artifact name collisions.

## Linked Issue

Closes #174

- [x] The linked issue is assigned to me.
- [x] This PR targets `develop`, or it is a release-tracked promotion into `master`.
- [x] This PR is not ready for review until the linked issue and test plan are complete.

## Package Impact

- [ ] Package format or manifest behavior
- [ ] Package discovery or enablement
- [x] Provider launch/shim behavior
- [ ] Setup or permission flow
- [ ] Documentation only
- [x] Tests only

## Safety

- [x] This change does not export secrets, credentials, private prompts, or machine-local state.
- [x] Package inputs are treated as untrusted.
- [x] Path handling prevents traversal outside the intended workspace/package root where applicable.
- [x] Provider-specific logic stays behind an explicit boundary.

## Verification

Relevant test cases:

- GitHub Actions Go test workflow now covers `ubuntu-latest` and `windows-latest`.
- Coverage artifacts are uploaded with OS-specific names.

```text
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/test.yml')); print('yaml ok')"
# yaml ok

go test ./...
# not run locally: Go toolchain is not installed in this WSL environment

go run ./cmd/lineage --help
# not run locally: Go toolchain is not installed in this WSL environment
```

If this PR does not need automated tests, explain why:

- This change modifies CI workflow configuration only. The PR CI run is the executable verification for the new Windows matrix leg.

## Release Tracking

Complete this section only for PRs targeting `master`.

- [ ] Release-worthy master promotion
- [x] Non-release housekeeping

Planned version:

`v0.0.0`

Release classification:

`patch | minor | major`

Release notes:

What changed:

-

What is experimental:

-

What is outside scope:

-

Safety and compatibility notes:

-

Post-merge tag plan:

- [ ] Create and push an annotated SemVer tag after merge to `master`.

Non-release housekeeping reason:

- CI coverage expansion for Windows-specific code paths.

## Notes For Reviewers

[...]

PR #212, opened 2026-08-24
fix: make Go test suite pass on Windows

### LINKED ISSUE #211: Windows CI exposes POSIX-only tests and path assumptions
## Problem

PR #209 adds `windows-latest` to the Go test workflow and uncovers existing Windows failures. The workflow change itself is small and correct, but the new Windows leg currently fails because several tests and code paths assume POSIX behavior.

## Failures observed

From the Windows job in `Test` run `32777771227`:

- `internal/atomicfile`
  - `TestWriteFileSetsPermissions`
  - Windows reports mode `-rw-rw-rw-` rather than the requested Unix permission bits.
- `internal/cli`
  - Multiple run/materialization tests use `/bin/echo`, which does not exist on Windows runners.
  - `TestRunPassesFlagsAfterDoubleDashToProvider` writes a `.sh` fake provider and tries to execute it as a Windows binary.
- `internal/materialize`
  - `TestApplyCapsStagedFilePermissions`
  - The test asserts Unix group/other write bits on Windows, where chmod semantics differ.
- `internal/packages`
  - `TestSafeJoinRejectsAbsolutePaths`
  - `TestPlanSetupRejectsUnsafePaths`
  - Tests use `/etc/passwd` as an absolute unsafe path; on Windows that is not an absolute path.
- `internal/provider`
  - `TestCandidateBinariesFindsMultipleAndSkipsShim`
  - `TestCandidateBinariesSkipsShimContentOutsideShimsDir`
  - Candidate-binary fixtures do not produce executable Windows candidates as written.

## Expected fix direction

Make the test suite genuinely cross-platform rather than weakening the new Windows CI signal:

- Split Unix permission-bit assertions behind platform-specific tests or expectations.
- Use OS-appropriate fake provider binaries/commands in CLI and provider tests.
- Add Windows absolute-path cases for `SafeJoin`/setup path safety.
- Keep the `windows-latest` CI leg once these failures are fixed.

## Context

This is not caused by the workflow-matrix change in #209. The new CI leg is exposing Windows compatibility gaps that already existed but were not exercised in CI.

priyam-jain-2002:
## Summary

- Split Unix-exact permission-bit assertions behind OS-specific helpers in `internal/atomicfile` and `internal/materialize`, since Windows `Chmod` models read-only/writable rather than full POSIX bits.
- Replaced POSIX-only absolute-path fixtures in `internal/packages` with paths built from `filepath.VolumeName`, so traversal rejection is exercised correctly on POSIX and Windows.
- Updated provider candidate-binary fixtures to use `.exe` on Windows, matching PATHEXT-based lookup.
- Replaced POSIX-only fake provider commands in `internal/cli` with OS-appropriate scripts where tests actually launch providers.

This does not change the CI workflow itself; #209 adds the `windows-latest` leg separately. This PR makes the existing test suite pass under that matrix.

## Linked Issue

Fixes #211

## Package Impact

- [ ] Documentation only
- [x] Tests only

## Safety

## Verification

Relevant test cases:

- Windows-specific permission assertions in `internal/atomicfile` and `internal/materialize`.
- Cross-platform absolute-path rejection tests in `internal/packages`.
- Windows candidate-binary lookup fixtures in `internal/provider`.
- OS-specific fake provider launch fixtures in `internal/cli`.

```text
go test ./...
GOOS=windows GOARCH=amd64 go build ./...
go vet ./...
gofmt -l
```

- Not applicable.

## Notes For Reviewers

- Review this together with #209: #209 expands the CI matrix, while this PR fixes the POSIX assumptions that blocked that matrix from passing.


internal/cli/regression_test.go
```diff
@@ -4,13 +4,36 @@ import (
...
 
+// writeFakeProviderScript creates a fake provider executable that records
+// the arguments it's invoked with to markerPath, using whichever command
+// interpreter the current OS actually has (POSIX shell vs. Windows batch),
+// and returns the script's path.
+func writeFakeProviderScript(t *testing.T, dir, markerPath string) string {
+	t.Helper()
+	if runtime.GOOS == "windows" {
+		path := filepath.Join(dir, "fake-claude.cmd")
+		script := "@echo off\r\necho %* > \"" + markerPath + "\"\r\n"
```
adkalhan: Could we include an argument containing spaces in this test? %* doesn't preserve argument boundaries the same way "$@" does on the Unix version, so this test may not catch a Windows-specific argument handling issue. For example, passing "hello world" would make sure the Windows test is actually checking the argument forwarding behavior.

internal/materialize/permissions_unix_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"os"
+	"testing"
+)
+
+// assertNoLooseWriteBits checks that copyFile's 0o755 cap actually stripped
+// the group/other write bits a loose source file (0o777) would otherwise
+// carry over - a check only POSIX permission bits can make.
+func assertNoLooseWriteBits(t *testing.T, mode os.FileMode) {
```
adkalhan, line 13: This makes the Windows version of this assertion a no-op. Is there a Windows-specific property we can check here instead? If there isn't an equivalent of the POSIX group/other write bits, could we make that limitation explicit in the calling test so it's clear that this behavior is only being verified on POSIX?

priyam-jain-2002: Addressed in `8b4d8c6`: I moved the Windows limitation into the calling test with an explicit `t.Skip`, since Windows `chmod` does not expose POSIX group/other write bits through `os.FileMode`. The Windows helper now fails if called, so the POSIX-only assertion cannot silently become a no-op.

internal/cli/regression_test.go
priyam-jain-2002: Addressed in `8b4d8c6`: the regression test now forwards both `--dry-run` and `hello world`, and the fake provider records one argument per line. On Windows the batch script uses `%~1` plus `shift` instead of `%*`, so the test now verifies that an argument containing spaces survives as a single provider argument.

adkalhan (approved):

PR #215, opened 2026-08-25
Add Auggie support and refactor for adapter testing

### LINKED ISSUE #35: Add Auggie provider adapter
## Goal

Add Auggie / Augment Code as a Lineage provider target.

## Why

Auggie is Augment Code's terminal-based coding agent with deep codebase context. It is relevant for users who want a provider with strong repository understanding and CLI-first interaction.

## Scope

- Define the Auggie adapter boundary.
- Detect or configure the `auggie` command where available.
- Decide which package assets can map to Auggie config, context, MCP, or launch behavior.
- Avoid packaging Augment/Auggie auth state or session tokens.
- Add dry-run/inspect output showing what Auggie support will do.

## Acceptance Criteria

- Auggie support is behind an explicit provider adapter.
- No Augment/Auggie credentials or local login state are packaged.
- Tests cover launch planning and materialization behavior that Lineage owns.
- Docs explain required local Auggie setup and current limitations.

AlessandroDiGioacchino:
## Summary
Add Auggie / Augment Code as a supported Lineage provider.

The adapter:
- resolves the `auggie` executable from `PATH` or `providers.auggie.binary`;
- stages packaged Agent Skills under `.augment/skills/`;
- writes Lineage-managed context to the project’s `AGENTS.md`;
- inherits generic launch planning, argument forwarding, shims, doctor diagnostics, approval gating, and idempotent materialization.

The implementation uses `.augment/skills/` instead of `.augment/rules/` because current Auggie releases support native `SKILL.md` Agent Skills there. Rules are a separate Augment concept.

Provider tests were generalized to cover every registered provider, making future adapters easier to add without duplicating test cases.

## Linked Issue
Closes #35 

## Package Impact
- [x] Documentation only
- [x] Tests only

## Merge Method

## Safety

## Verification
Relevant test cases:
- Provider registry contracts for Claude, Codex, and Auggie.
- Configured binary resolution and launch arguments for every provider.
- Materialization of skills, context, and provider-specific state for every provider.
- Shim installation for every registered provider.
- CLI help and unknown-provider output derived from the provider registry.
- Dry-run launch planning for every provider.

```text
git diff --check
go test ./...
```
All tests pass.

## Notes For Reviewers
Please review the Auggie filesystem mapping in particular:
- Skills: .augment/skills/
- Context: AGENTS.md

Auggie requires Node.js 22 or later, and local authentication through `auggie login`. Also, we should keep in mind that Auggie is beta.
Please take a look at the tests to determine if I've generalized them correctly. While I've only added Auggie as provider, the way tests are written now should allow to easily add other providers.
Finally, I've edited the docs to include Auggie: it works for now, but depending on the number of providers it could require some reformat.


priyam-jain-2002 (changes requested):
Thanks for taking this on. The provider-registry direction is a good fit for Lineage, and the broadening of the provider tests is the right shape for keeping future adapters from reintroducing Claude/Codex hardcoding.

I’m going to request changes before merging, mainly because the Auggie materialization path does not yet appear to guarantee that the staged skills are valid Auggie skills.

The key issue is that this registers Auggie as a native skills provider at `.augment/skills/`, then the existing materializer copies each package skill directory unchanged. Lineage package discovery currently treats any `skills/<name>/SKILL.md` as a skill, including files with no YAML frontmatter; several tests intentionally build packages with just `# Review` / `# Hello`. Auggie’s documented native skill format is stricter: each `SKILL.md` needs required frontmatter (`name`, `description`), and the `name` must match the directory name. Since Lineage stages skills as `<package>-<skill>`, a normal package skill with no frontmatter, or with `name: <skill>`, can be copied into `.augment/skills/<package>-<skill>/SKILL.md` but not load as the Auggie skill this adapter promises.

Please either:

- render Auggie-specific skill files so the staged `SKILL.md` has valid frontmatter and a `name` matching the staged directory, with tests covering a plain Lineage skill and an already-frontmattered skill; or
- choose a non-native Auggie integration path for package content and document that limitation clearly.

There are also a couple of smaller cleanup items:

- `gofmt -l` still reports `internal/cli/cli_test.go`, `internal/materialize/materialize_test.go`, `internal/provider/provider_test.go`, and `internal/provider/registry_test.go`.
- `TestDoctorReportsEachKnownProvider` still checks only `claude` and `codex`; since this PR claims Auggie inherits doctor diagnostics, that test should iterate over `provider.Known()` like the other generalized provider tests.

Local checks after merging current `develop` into the PR branch:

- `go test ./...` passes
- `go vet ./...` passes
- `go build ./...` passes
- `git diff --check origin/develop...tmp-pr-215-review` passes
- formatting check fails on the four files listed above

The branch is currently behind `develop`, but it merged cleanly in my local review checkout. Once the Auggie skill-loading contract and formatting are addressed, this should be in a much better place for another pass.

internal/provider/registry.go
```diff
@@ -22,6 +22,7 @@ type Provider struct {
 var registry = []Provider{
 	{Name: "claude", SkillsDir: filepath.Join(".claude", "skills"), ContextFile: "CLAUDE.md"},
 	{Name: "codex", SkillsDir: filepath.Join(".agents", "skills"), ContextFile: "AGENTS.md"},
+	{Name: "auggie", SkillsDir: filepath.Join(".augment", "skills"), ContextFile: "AGENTS.md"},
```
priyam-jain-2002: This is the right central place to register Auggie, but using `.augment/skills` makes Auggie a native skill consumer rather than just another copy target. Auggie requires each staged `SKILL.md` to have `name`/`description` frontmatter and the `name` must match this staged directory. Lineage currently accepts plain `SKILL.md` files, so this adapter needs either provider-specific rendering/validation before we can treat `.augment/skills` as a working contract.

internal/materialize/materialize_test.go
```diff
@@ -82,30 +82,69 @@ func TestApplyRejectsCollidingSkillDirNames(t *testing.T) {
...
-		t.Fatalf("read CLAUDE.md: %v", err)
-	}
-	content := string(contextData)
-	if !containsAll(content, beginMarker, endMarker, "review-pack@0.1.0") {
-		t.Fatalf("CLAUDE.md missing expected content:\n%s", content)
+	for _, adapter := range provider.Known() {
+		t.Run(adapter.Name, func(t *testing.T) {
+			root := t.TempDir()
+
+			if err := Apply(root, adapter, []packages.Package{pkg}); err != nil {
```
priyam-jain-2002, line 286: This broad provider test is useful, but for Auggie it currently only proves the file was copied. Please add an Auggie-specific assertion that the staged `SKILL.md` is actually loadable by Auggie: required frontmatter is present, `name` matches `review-pack-review`, and a description is available. A plain Lineage skill like this fixture is the important regression case.

README.md
```diff
@@ -213,13 +216,35 @@ lineage list
...
 
 lineage install-shims
 lineage doctor
 lineage version
 ```
 
+### Auggie setup and limitations
+
+Auggie is installed and authenticated separately on the receiving machine. It
+currently requires Node.js 22 or later:
```
priyam-jain-2002, line 275: The current Augment install docs list Node.js 20+ as the system requirement. Please align this setup text, or mention why Lineage is intentionally requiring a stricter version. The same correction should be mirrored in the sharing guide where the Auggie setup text is repeated.

internal/cli/cli_test.go
```diff
@@ -45,21 +46,41 @@ func TestEnableAndDryRun(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	cfg.Providers = map[string]config.Provider{"codex": {Binary: "/bin/echo"}}
+
+	cfg.Providers = map[string]config.Provider{}
+
+	for _, adapter := range provider.Known() {
+		cfg.Providers[ adapter.Name ] = config.Provider{
```
priyam-jain-2002: Please run `gofmt` on the changed Go test files. This line is one example; `gofmt -l` reports `internal/cli/cli_test.go`, `internal/materialize/materialize_test.go`, `internal/provider/provider_test.go`, and `internal/provider/registry_test.go`.

internal/materialize/materialize_test.go
AlessandroDiGioacchino: I've added a couple of tests instead of editing this one. Namely, `TestApplyRendersPlainSkillForAuggie` and `TestApplyUpdatesExistingFrontmatterForAuggie`

AlessandroDiGioacchino:
Okay, let me know what you think. Once again, I tried considering that we want to add other adapters as well, so I used the Strategy pattern (mainly contained in `skill_renderer.go`). `materialize` should be able to delegate the work to the specific provider.

priyam-jain-2002 (approved):
Thanks for the follow-up here. I reviewed the current head (`25e6082`) against the earlier discussion.

The previous blockers look addressed to me:

- Auggie now has provider-specific skill rendering, so plain Lineage `SKILL.md` files are staged with required frontmatter and `name` matching the staged directory.
- Existing Auggie/frontmatter metadata keeps its description while the staged name is corrected.
- The materialization tests now cover both the plain-skill and already-frontmatter cases for Auggie.
- `doctor` now reports every provider from `provider.Known()`.
- The repeated Node.js setup text is aligned to Node.js 20+.
- The earlier formatting issue is gone.

Local verification:

```text
go test ./...
go vet ./...
gofmt -l internal/cli/cli_test.go internal/materialize/materialize_test.go internal/provider/provider_test.go internal/provider/registry_test.go internal/provider/skill_renderer.go internal/shim/shim_test.go
git diff --check origin/develop...HEAD
GOCACHE=$PWD/.tmp-go-cache GOMODCACHE=$PWD/.tmp-go-modcache GOOS=windows GOARCH=amd64 go build ./...
```

All passed. I also resolved the now-addressed review threads. From my side this is ready once GitHub has normal CI signal on the current head.

internal/materialize/materialize.go
```diff
@@ -124,6 +125,25 @@ func apply(projectRoot string, adapter provider.Provider, pkgs []packages.Packag
...
 		}
+
+		skillPath := filepath.Join(dest, "SKILL.md")
+		source, err := os.ReadFile(skillPath)
+		if err != nil {
+			return fmt.Errorf("read staged skill %s: %w", rel, err)
+		}
+
+		stagedName := filepath.Base(dest)
+		rendered, err := adapter.RenderSkill(stagedName, source)
```
priyam-jain-2002: The provider-owned rendering boundary looks right to me. One implementation edge case I want fixed before merge: `Apply` copies the skill directory into the provider location before calling `adapter.RenderSkill`. If Auggie rendering then fails, for example because a `SKILL.md` starts with `---` but has invalid/unterminated frontmatter, we return an error and avoid launching, but the raw copied skill may already remain under `.augment/skills`.

Can we render/validate the `SKILL.md` before replacing the existing staged directory, or otherwise make this failure path avoid leaving an invalid Auggie skill staged?

priyam-jain-2002 (changes requested):
Thanks for the follow-up here. The architectural direction looks right to me: Lineage’s package format stays provider-neutral, and provider-specific projection belongs at the adapter boundary. I agree with keeping this PR scoped to Auggie support plus the minimal provider-owned rendering hook needed to make Auggie materialization correct, rather than expanding into Claude/Codex migration or a broader provider-interface redesign.

I am requesting one small implementation fix before merge: the materialization failure path should not leave a raw, invalid Auggie skill staged if provider rendering fails after the copy. The inline comment has the concrete case.

Maintainer-side polish, like PR description consistency, can be handled separately.


Small process note: sorry for the noisy review sequence here. An automated review/helper flow briefly posted an approval before I finished reconciling the follow-up implementation concern. This latest review is the intended current state: architecture looks good, but I’d like the render-after-copy failure path fixed before merge.

priyam-jain-2002:
Thanks for the latest update. I re-reviewed the current PR head (`ee7de0b`) against the remaining concerns.

The Auggie materialization fix looks good to me now: `Apply` renders/validates the desired skills before replacing the existing staged directories, and the regression coverage confirms a failed Auggie render leaves the previously staged skill untouched.

The only thing I would still treat as blocking merge is that the branch currently conflicts with current `develop` in:

- `internal/cli/cli_test.go`
- `internal/materialize/materialize_test.go`

Those look like test-side conflicts from recent `develop` changes rather than a design issue in this PR.

Two small non-blocking follow-ups I noticed:

- Augment's current public install docs/npm package now say Auggie requires Node.js 22+, while this branch says 20+. Since that requirement appears to have drifted during review, I would not block this PR on it.
- The Auggie skill renderer only recognizes LF frontmatter delimiters (`---\n` / `\n---\n`), so a CRLF-authored `SKILL.md` with existing frontmatter may keep the old frontmatter in the body instead of preserving its metadata cleanly. The generated staged skill still gets valid Auggie `name`/`description` frontmatter, so I'd treat this as a robustness polish item rather than a merge blocker.

Once the branch is updated cleanly against `develop` and CI is green, I'm comfortable with this from my side.

internal/materialize/materialize.go
priyam-jain-2002: This looks addressed in the current head. `Apply` now prepares each desired skill first: it reads the source `SKILL.md`, calls `adapter.RenderSkill(...)`, and returns on render error before removing or copying any staged skill directory. The added regression test also covers the failure path by confirming an invalid Auggie frontmatter update leaves the previously staged skill unchanged.

Resolving this thread from my side.

docs/guides/share-agent-workflows.md
```diff
@@ -85,7 +88,13 @@ Before enabling, receivers can review:
 - `lineage.yaml` manifest metadata, exports, entrypoints, and capabilities.
 - `skills/`, `workflows/`, `agents/`, `policies/`, and `references/` content.
 - Validation output, including portability and safety findings.
-- Dry-run materialization output for Claude Code or Codex.
+- Dry-run materialization output for Claude Code, Codex, or Auggie.
+
```
priyam-jain-2002, line 92: This says Node.js 20+, but the PR description and README addition say Auggie currently requires Node.js 22+. Please use the verified requirement consistently; otherwise a receiver following this guide can satisfy our documented prerequisite and still fail to install/run Auggie.

priyam-jain-2002 (changes requested):
The adapter direction is consistent with the provider boundary and the generalized tests are useful, but there is one documentation correctness issue to fix before merge. The PR itself and README state Auggie requires Node.js 22+, while `docs/guides/share-agent-workflows.md` says Node.js 20+. We should not publish contradictory receiver prerequisites. Please make the guide match the verified/current requirement and keep the requirement consistent across all Auggie docs.

priyam-jain-2002 (approved):

priyam-jain-2002:
Thank you for sticking with this one through the full review cycle. This ended up being a fairly long PR, and I appreciate the care you took in responding to the feedback rather than just patching around it — especially around provider-specific skill rendering, failure-safe materialization, the generalized adapter tests, and the documentation follow-ups.

The latest commit addresses the remaining Node.js requirement inconsistency, both required checks are green, and the outstanding review thread is resolved. I’ve approved the current head.

Thanks again for the contribution and for being patient through the iterations — this is a meaningful addition to Lineage.

AlessandroDiGioacchino:
My pleasure 👍️ Sorry if the latest commits took longer, I've been busy with work and stuff.

PR #219, opened 2026-08-27
doctor: validate materialize state, graph.json, and snapshot integrity

### LINKED ISSUE #200: Extend lineage doctor to validate materialize state, graph.json, and snapshot integrity
## Problem

ADR 0015 (`docs/decisions/0015-the-lineage-directory-is-a-versioned-enumerated-container.md`, from #59) enumerates every file `lineage` writes under project- and user-level `.lineage/` and gives `lineage doctor` a real spec to validate against for the first time. Today `doctor` only checks:

- project `config.yaml` validity (now including its new `schema` field, transitively, since `LoadProjectConfig` rejects an unsupported schema)
- enabled-package resolution
- shim PATH placement
- provider binary resolution

It does not check:

- `materialized-<provider>.json` schema/staleness (e.g. a state file referencing skill dirs that no longer exist on disk)
- `.lineage/graph.json` parseability/integrity
- snapshot store referential integrity — whether every object a snapshot manifest under `~/.lineage/snapshots/` references actually exists under `~/.lineage/objects/`

## Ask

Extend `lineage doctor`'s scope to cover the above, using ADR 0015's regenerable-vs-authoritative classification to decide severity: a broken *regenerable* file (materialize state) can likely be a warning with a suggested fix (re-run `lineage run`), while a broken *authoritative* file (`graph.json`, a snapshot manifest with a missing object) should probably be a harder failure, since there's no automatic way to regenerate lost history.

Design questions worth resolving as part of this:

- Warning vs. fatal for each new check, consistent with `doctor`'s existing "fails only for things that are actually broken" posture.
- Whether `doctor` should ever offer to repair anything itself, or stay read-only/diagnostic-only as it is today.

## Non-goals

Not about adding new `.lineage` entries or changing existing formats — this is purely extending `doctor`'s validation coverage to the spec ADR 0015 already wrote down.

priyam-jain-2002:
## Summary

Extends `lineage doctor` to validate the remaining structured state identified by ADR 0015 and closes integrity gaps found during review:

- Warn when recorded provider skill directories are missing, regular files, or symlinks, since materialization state is regenerable.
- Fail when `.lineage/graph.json` is unreadable.
- Fail when the snapshot store has malformed layout, malformed manifests, or missing/corrupt referenced objects.
- Validate snapshot schema, package identity, canonical unique file paths, required `lineage.yaml`, and strict SHA-256 object IDs before a snapshot is accepted or materialized.
- Refuse symlinked/non-regular content-addressed objects and verify an existing deduplicated object before reusing it.

`doctor` remains diagnostic-only and does not mutate or auto-repair state.

## Linked Issue

Fixes #200.

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- Materialize state is healthy immediately after apply.
- Missing, file-replaced, and symlink-replaced staged skill directories are reported as stale.
- Missing materialize state remains a clean no-op.
- Corrupt graph state is fatal.
- Clean snapshot stores and missing/corrupt referenced objects are distinguished.
- Self-hashed but structurally invalid snapshot manifests are rejected.
- Non-canonical object IDs and malformed store entries are rejected.
- Snapshot creation remains deterministic and deduplicated.
- Snapshot restoration rejects invalid metadata before writing.

```text
go test ./...
go vet ./...
go build ./...
```

## Notes For Reviewers

Please scrutinize the boundary choices: materialize-state defects remain warnings because rerunning materialization repairs them; graph and snapshot-store defects remain fatal because those stores may be the only authoritative provenance/content copy. Symlinked staged skills are intentionally treated as stale because Lineage itself creates real directories and following a replacement symlink could expose provider discovery to content outside the project.

The branch is merged with current `develop`; conflict resolution retained the current release notes, provider additions, and updated test infrastructure.


internal/materialize/materialize.go
```diff
@@ -207,6 +207,37 @@ func equalStrings(a, b []string) bool {
...
+// materialized (no state file yet) or every recorded skill dir is present.
+func DiagnoseState(projectRoot, providerName string) ([]string, error) {
+	s, err := loadState(projectRoot, providerName)
+	if err != nil {
+		return nil, err
+	}
+	var missing []string
+	for _, rel := range s.SkillDirs {
+		if _, statErr := os.Stat(filepath.Join(projectRoot, rel)); os.IsNotExist(statErr) {
+			missing = append(missing, rel)
```
adkalhan, line 298: P2: A recorded skill directory can be replaced by a regular file and still pass this check, because `os.Stat` succeeds. `lineage doctor` then reports healthy materialize state even though provider discovery will be broken.

Please require the path to be a directory (`info.IsDir()`); consider `Lstat` too if symlinks should not count as a staged skill directory. Please add a regression test for a staged skill directory replaced by a file.

internal/cli/cli.go
```diff
@@ -1405,6 +1441,32 @@ func runDoctor(home string, stdout, stderr io.Writer) error {
 		}
 	}
 
+	if manifestIDs, err := snapshot.AllManifestIDs(home); err != nil {
+		fmt.Fprintf(stdout, "snapshot store: FAIL - could not list snapshot manifests: %v\n", err)
+		broken = true
+	} else {
+		snapshotProblems := 0
+		for _, id := range manifestIDs {
+			m, err := snapshot.LoadManifest(home, id)
```
adkalhan, line 1462: P2: `LoadManifest` currently only JSON-decodes the stored object. A self-hashed but structurally invalid snapshot manifest—such as `null` or `{"schema":99}`—decodes to a zero-value manifest. Because it has no files, the loop below verifies nothing and `doctor` reports the snapshot store as OK.

Please validate the manifest’s schema, required identity fields, and file entries before treating it as healthy, and add coverage for malformed-but-valid-JSON manifests.

adkalhan (dismissed):
## Changes requested

Two integrity checks can currently report a healthy state when the underlying runtime state is unusable:

1. Invalid but self-hashed snapshot manifests can pass `lineage doctor`.
2. A staged skill directory replaced with a regular file can pass materialize-state validation.

Local validation passed: `go test ./...`, `go vet ./...`, and `go build ./...`.

internal/materialize/materialize.go
priyam-jain-2002: Addressed. `DiagnoseState` now uses `os.Lstat` and requires a real directory, so both a regular-file replacement and a symlink replacement are reported as stale. I added regressions for both cases; symlinks are intentionally rejected because Lineage materializes real directories and following one could expose provider discovery to content outside the project.

internal/cli/cli.go
priyam-jain-2002: Addressed. `LoadManifest` now rejects `null`, unsupported/explicit-zero schemas, missing or invalid package identity, empty/malformed file entries, unsafe/non-canonical or duplicate paths, missing `lineage.yaml`, unsorted paths, and non-canonical SHA-256 object IDs. `Materialize` validates the same invariant set before writing. The table-driven regression suite includes the reported `null`/schema cases plus adjacent structural bypasses.

priyam-jain-2002:
## Follow-up review finding

**P2 — validate recorded `skill_dirs` before joining them to the project root.**

`DiagnoseState` treats `.lineage/materialized-*.json` as trusted and calls `os.Lstat(filepath.Join(projectRoot, rel))`. A corrupted or manually edited state file can use `../` or an absolute path, making `lineage doctor` probe outside the project and report that external path as healthy/stale.

Please require each recorded path to be a clean, non-absolute, project-contained path (and ideally under that provider’s `SkillsDir`) before `Lstat`; invalid state should become the existing regenerable warning. Add regressions for traversal and absolute entries.

Rushilkr77 (changes requested):
Reviewed against current `develop`. Good progress: both of adkalhan's original findings are fixed and verified (`DiagnoseState` now `Lstat`s and requires a real directory — file/symlink replacement both flagged stale; `LoadManifest` now rejects `null`/bad-schema/malformed snapshot manifests with regressions for each).

Two things still block merge:

1. **Unaddressed path-traversal finding (my 2026-09-13 comment on `internal/materialize/materialize.go:296`)** — `DiagnoseState` still does `os.Lstat(filepath.Join(projectRoot, rel))` with `rel` taken straight from `.lineage/materialized-*.json`, untrusted per this PR's own stated threat model. A `rel` of `../../etc` or an absolute path makes `doctor` probe outside the project. No commit has landed since that comment (last commit is 2026-09-06, comment is 2026-09-13). Please reject non-clean/absolute/`..`-containing entries (surface as the existing regenerable warning) before `Lstat`, and add traversal + absolute-path regressions.
2. **Merge conflicts with `develop`** — GitHub reports this branch as conflicting/dirty right now. Please rebase.

Once both are resolved I'll do a final pass on the snapshot-store integrity checks (schema/identity/canonical-path/object-ID validation) which look solid on this read-through.

priyam-jain-2002:
Author update: I rechecked the latest branch state and the remaining follow-up is the recorded `skill_dirs` path-containment finding. That is still valid on the current head: `DiagnoseState` joins state-file values directly to the project root without first rejecting absolute/traversal paths.

I’m going to address that on this branch with explicit clean/relative/project-contained validation plus traversal and absolute-path regressions before asking for another review. I’m not treating the older green CI as sufficient for this newer finding.

Rushilkr77 (changes requested):
Re-checked the current head (`85df22b`) against my 2026-09-13 comment on `internal/materialize/materialize.go:296`. It's still open.

`DiagnoseState` does:

```go
info, statErr := os.Lstat(filepath.Join(projectRoot, rel))
```

with `rel` read straight from `.lineage/materialized-*.json` — untrusted per this PR's own threat model. A `rel` of `../../etc` or an absolute path makes `lineage doctor` probe outside the project. Both of adkalhan's original findings (self-hashed snapshot manifests, file/symlink replacing a staged skill dir) are genuinely fixed and regression-tested. This one isn't — no commit has landed addressing it since `d48bba1`, only merge commits after your 09-16 comment saying you'd fix it on this branch.

Please:
1. Reject non-clean, absolute, or `..`-containing `rel` entries before the `Lstat` (surface as the existing regenerable warning, same as a missing/replaced dir).
2. Add traversal and absolute-path regression tests alongside the existing symlink/file-replacement ones.

`go build`, `go vet`, and `go test ./internal/materialize/... ./internal/cli/... ./internal/graph/... ./internal/snapshot/...` all pass otherwise. Once the path-containment fix lands I'll do the final pass on the snapshot-store integrity checks.

Rushilkr77 (approved):
Path-containment fix landed in \`7b08f68\`. Verified on current head:

\`recordedSkillDirPath\` rejects empty/absolute/non-clean \`rel\`, walks components for \`..\`, requires containment under the adapter's \`SkillsDir\`, and does a final \`filepath.Abs\`/\`filepath.Rel\` containment check as a backstop (covers the case where the component loop splits on \`os.PathSeparator\` but a stored path uses forward slashes on Windows — the final check catches it regardless). Both requested regressions are present: \`TestDiagnoseStateReportsTraversalSkillDirEntry\`, \`TestDiagnoseStateReportsAbsoluteSkillDirEntry\`.

\`go build\`, \`go vet\`, and \`go test ./internal/materialize/... ./internal/cli/... ./internal/graph/... ./internal/snapshot/...\` all pass on this head.

Also did the promised final pass on the snapshot-store integrity checks (schema/identity/canonical-path/object-ID validation) — solid, no further findings.

Approving.

priyam-jain-2002:
Addressed the remaining recorded `skill_dirs` path-validation finding in commit 7b08f68.

`DiagnoseState` now rejects empty, non-clean, absolute, and `..`-containing entries before `Lstat`. Recorded paths must remain within the project and under the selected provider's `SkillsDir`. Invalid entries become the existing regenerable materialize-state warning.

Added traversal and absolute-path regression tests using existing external directories, so those entries cannot be reported as healthy.

Validation passed: `go test ./...`, `go vet ./...`, and `go build ./...` (using a writable module cache). The focused materialize tests and `git diff --check` also passed before push. The branch includes current `develop` (993e9f8), resolving the earlier conflict concern.

Ready for the final snapshot-store integrity review.

PR #222, opened 2026-08-27
Add Windsurf provider adapter

### LINKED ISSUE #34: Add Windsurf provider adapter
## Goal

Add Windsurf as a Lineage provider target.

## Why

Windsurf is a widely used AI coding environment and should be considered for Lineage package activation where local config or workspace assets can be materialized safely.

## Scope

- Define the Windsurf adapter boundary.
- Determine whether Lineage should support command launch, editor/workspace materialization, or both.
- Map supported package assets to Windsurf-compatible local files.
- Keep Windsurf-specific behavior outside core package semantics.
- Add dry-run/inspect output showing what Windsurf support will do.

## Acceptance Criteria

- Windsurf support is behind an explicit provider adapter.
- Receiver-side writes are visible before apply.
- Tests cover supported materialization behavior.
- Docs explain what is supported and what must still be configured by the user.

alexsmolya:
This adds the project-scoped Windsurf provider adapter.

- materializes skills under `.windsurf/rules/`
- writes the generated context to the legacy-compatible project-local `.windsurfrules`
- keeps Windsurf config/materialization-only: no GUI command launch and no global MCP configuration
- adds registry, dry-run, materialization, shim, and documentation coverage

Relevant test cases:

- `TestWorkflowRunWindsurfMaterializesWithoutLaunching` covers workflow materialization without launching Windsurf.
- `TestApplyStagesSkillsForWindsurfAdapter` covers Windsurf skill materialization.
- `TestResolveMaterializeOnlyProviderDoesNotResolveBinary` covers skipping binary resolution for the materialization-only capability.
- `TestGetWindsurfProvider` and `TestBuildPlanWindsurfDryRun` cover provider registration and dry-run planning.
- `TestInstallWritesShimsOnlyForLaunchableProviders` covers excluding Windsurf from shim installation.

Validation: `gofmt`, `git diff --check`, `go test ./...`, `go vet ./...`, `make test`, and `make build`.

Closes #34

## Merge Method

## Verification

```text
go test ./...
```


priyam-jain-2002 (changes requested):
Thanks for the Windsurf adapter work. The materialization target itself looks reasonable, but I think this needs changes before merge because the current implementation makes Windsurf behave like a launchable/shimmed provider even though the PR describes it as materialization-only.

1. `lineage run windsurf` still launches the configured/resolved binary after materialization.

   Adding `windsurf` to `internal/provider/registry.go` automatically sends it through the generic `runProvider` path, which calls `provider.Launch(plan.ProviderPlan)` after `materialize.Apply(...)`. `workflow run ... windsurf` does the same after `ApplyWorkflow(...)`.

   That conflicts with the PR description and README language saying Windsurf is config/materialization-only and does not launch the Windsurf GUI. This should either be represented as provider capability metadata, or otherwise enforced so Windsurf materializes project files and exits without resolving/running a Windsurf binary.

2. `install-shims` now installs a `windsurf` shim.

   `shim.Install` iterates `provider.Known()`, so registering Windsurf creates a shim that executes `lineage run windsurf "$@"`. That means putting the shim directory on PATH can intercept normal `windsurf` invocations and route them through Lineage, again contradicting “no GUI command launch.” If Windsurf is materialization-only, it should not get a shim by default.

3. Help/docs are inconsistent around this behavior.

   The top-level help still says `install-shims` puts Lineage in front of `claude/codex`, while the code now includes every provider. The README also has one place where workflow dry-run still lists only `<claude|codex>` and another where it includes Windsurf. These should be aligned once the intended Windsurf behavior is enforced.

Validation I ran locally: `go test ./...` and `git diff --check`, both passing on the current PR branch.

README.md
```diff
@@ -233,13 +235,19 @@ lineage list
...
+lineage workflow run <workflow-name> <claude|codex|windsurf> [--dry-run] [--yes]
 
 lineage install-shims
 lineage doctor
 lineage version
 ```
 
+Windsurf support is project-scoped and config/materialization-only. Lineage
+stages skills under `.windsurf/rules/` and writes its generated context to the
+legacy-compatible project-local `.windsurfrules` file. It does not launch the
```
priyam-jain-2002, line 253: This docs statement says Lineage does not launch the Windsurf GUI, but the implementation still calls the generic launch path after materialization. Please make the runtime behavior match this statement before merging; otherwise users who configure or discover a `windsurf` binary will see Lineage try to execute it.

internal/shim/shim_test.go
```diff
@@ -29,10 +31,10 @@ func TestInstallWritesShimForEveryKnownProvider(t *testing.T) {
 		t.Fatalf("Install() error = %v", err)
 	}
 
-	for _, name := range []string{"claude", "codex"} {
-		path := filepath.Join(home, "bin", FileName(name, runtime.GOOS))
+	for _, p := range provider.Known() {
```
priyam-jain-2002, line 34: Because this test now expects shims for every `provider.Known()` entry, the new Windsurf provider will get a `windsurf` shim from `install-shims`. That shim intercepts normal `windsurf` invocations and routes them through `lineage run windsurf`, which contradicts the “no GUI command launch”/materialization-only design. Windsurf should probably be excluded from shim generation, with test coverage for that exclusion.

internal/provider/registry.go
```diff
@@ -22,6 +22,7 @@ type Provider struct {
 var registry = []Provider{
 	{Name: "claude", SkillsDir: filepath.Join(".claude", "skills"), ContextFile: "CLAUDE.md"},
 	{Name: "codex", SkillsDir: filepath.Join(".agents", "skills"), ContextFile: "AGENTS.md"},
+	{Name: "windsurf", SkillsDir: filepath.Join(".windsurf", "rules"), ContextFile: ".windsurfrules"},
```
priyam-jain-2002: This single registry entry makes Windsurf participate in the same generic provider paths as Claude/Codex. In particular, `lineage run windsurf` will still build a provider plan and call `provider.Launch` after materialization, which conflicts with the PR contract that Windsurf is config/materialization-only and should not launch the GUI. Please encode that capability difference in the provider model or otherwise skip binary resolution/launch for Windsurf.

priyam-jain-2002 (approved):
Re-reviewed the latest head. The earlier provider-boundary blocker is addressed cleanly: Windsurf is now represented as materialization-only, binary resolution/launch is skipped for that capability, shim installation filters it out, and the CLI/docs/tests now describe the same behavior. The new regression coverage also exercises both plain `run` and workflow materialization without launch.

Code-wise this is good to merge from my side. GitHub is still reporting the PR as not mergeable and the current Actions runs are `action_required`, so those repository/CI gates still need to clear before merge.

README.md
```diff
@@ -203,6 +208,8 @@ providers:
     binary: /path/to/real/claude
   codex:
```
priyam-jain-2002, line 213: This example still asks users to configure `providers.windsurf.binary`, but the implementation now marks Windsurf as materialization-only and returns before binary resolution/launch. That makes this config entry misleading and inconsistent with the surrounding docs that say `lineage run windsurf` works without resolving a Windsurf binary. Please remove the Windsurf binary example here, or explicitly state that current Windsurf support does not use it.

priyam-jain-2002 (changes requested):
Re-reviewed the latest head after resolving the stale launch/shim threads that are now addressed in code. There is one remaining docs inconsistency before merge: the README still documents a Windsurf binary config entry even though Windsurf is now explicitly materialization-only and `Resolve` skips binary lookup for that provider. Please remove that config example or clearly mark it as not used by current Windsurf support.

priyam-jain-2002 (dismissed):
Re-reviewed the latest head. The Windsurf binary config example was removed, and the previous materialization-only, launch-skip, shim-exclusion, and docs concerns are addressed. I do not see any remaining code-review blockers in the changed files.

GitHub still reports this PR as `mergeable_state: dirty`, so please resolve the merge conflicts against `develop` before merge. No merge/close action from me here.

priyam-jain-2002 (approved):

PR #225, opened 2026-08-29
test: encode legacy materialize state fixtures

priyam-jain-2002:
## Summary

Maintainer housekeeping follow-up for the Windows CI rollout in PR #209.

The `TestNeedsApprovalDefaultsMissingSchemaToCurrent` fixture was manually building schema-less JSON with `filepath.ToSlash(rel)`. That made the test fail on Windows because `NeedsApproval` computes platform-native relative skill directories with `filepath.Join`, so the loaded legacy state and desired state differed only by separator style.

This changes the fixture to use `json.Marshal` with the platform-native `rel` value. The state remains schema-less, so the test still verifies the intended backwards-compatibility path; it just writes valid JSON without changing path semantics.

## Linked Issue

Maintainer housekeeping follow-up for PR #209 and PR #212. No standalone issue; this is the narrow Windows test blocker exposed by the Windows CI matrix.

- [x] The linked issue is assigned to me, or this PR is maintainer follow-up housekeeping.
- [x] This PR targets develop, or it is a release-tracked promotion into master.
- [x] This PR is ready for review after the linked context and test plan are complete.

## Package Impact

- [ ] Documentation only
- [x] Tests only

## Safety

## Verification

Relevant test cases:

- `TestNeedsApprovalDefaultsMissingSchemaToCurrent` now writes schema-less state through JSON encoding while preserving platform-native path separators.

```text
go test ./internal/materialize
go test ./...
```

Both pass locally on macOS.

- N/A; this PR changes a test fixture and was verified with the affected package plus full suite.

## Notes For Reviewers

This is intentionally test-only. It does not change `NeedsApproval` or materialized state behavior; it only fixes the legacy-state fixture so Windows does not compare `.claude/skills/...` from the hand-built JSON against `.claude\skills\...` from `filepath.Join`.

### DIFF
--- a/internal/materialize/materialize_test.go
+++ b/internal/materialize/materialize_test.go
@@ -1,6 +1,7 @@
 package materialize
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


priyam-jain-2002:
Verification note: I created temporary draft PR #226 combining PR #209 (Windows matrix) with this PR head to test the exact scenario that local macOS cannot execute. The combined tree passed PR Policy, Go (ubuntu-latest), and Go (windows-latest). This confirms the materialize legacy-state fixture fix clears the pasted Windows failure before merging this PR.

adkalhan (approved):

priyam-jain-2002:
Addressed the review/CI concern in `860b7c7` by normalizing materialized state `skill_dirs` when loading state, while keeping the regression fixture in legacy slash-encoded form. Verified locally with `go test ./...`.

Issue #128, opened 2026-08-22
Safety compliance: add package instruction-risk scanning


priyam-jain-2002:
## Goal

Scan package instructions for risky agent behavior before publish, inspect, enable, and execution.

## Why this matters

A Lineage package can influence an agent through skills, workflows, agents, policies, references, and provider-specific files. The most dangerous content may be an instruction, not a credential.

## Pipeline placement

- Publish pipeline: block or warn on risky instruction patterns before registry upload.
- Inspect pipeline: show instruction-risk findings by file.
- Enable pipeline: require explicit user confirmation for warning-level risky instructions.
- Execution/runtime pipeline: future enforcement can prevent undeclared risky behavior.

## Examples to flag

- "Ignore previous instructions" or equivalent prompt override language.
- "Read all files", "scan the home directory", or broad local exfiltration prompts.
- "Send/curl/upload secrets" or network exfiltration instructions.
- "Disable safety checks" or "do not ask the user".
- Silent destructive operations, such as deleting files without confirmation.
- Instructions that ask the agent to collect passwords, tokens, private keys, or browser session data.

## Strategy

Start with a small explainable rule set over package-controlled instruction files: `SKILL.md`, `WORKFLOW.md`, agents, policies, adapters, and setup templates. Report file path, category, severity, and explanation.

Do not pretend this is complete prompt-injection defense. Treat it as a practical risk signal that catches obvious malicious or sloppy packages.

## Decision needed

Decide which patterns are hard stops, which are warnings, and whether package authors can annotate/justify expected risky instructions.

## Acceptance criteria

- Instruction-risk scanner runs during validate/publish and is visible in inspect.
- Findings are categorized and do not rely on vague raw grep output.
- Tests include malicious examples and benign false-positive guardrails.
- Enable plan highlights warning-level findings before materialization.

## Out of scope

- Full LLM-based prompt safety review.
- Runtime sandbox enforcement.

Raghavrajvanshi:
### Proposal: instruction-risk scanner for publish/inspect/enable/execution

I'd like to take this one. Posting my understanding and plan first, per the issue-ownership process, before starting.

**Gaps this closes (checked against current code):**
- `Publish`/`Validate` only run the existing secret scanner (`internal/packages/secrets.go`) — no instruction-risk equivalent exists yet.
- `runInspect` never calls `Validate` at all, so findings aren't visible there today, even the existing secret ones.
[...]

PR #228, opened 2026-08-31
Risk scanner

Raghavrajvanshi:
## Summary

Adds an instruction-risk scanner (`ScanForInstructionRisk`) that flags packages whose skill/setup content contains risky-instruction patterns — prompt-override attempts, broad local file reads, silent destructive operations, credential collection, and network exfiltration of credential-shaped data — and surfaces the findings through `validate`/`inspect`/`enable`/`add`. This is the scanner half of #128's safety boundary.

This branch also addresses the review rounds so far: it unifies excerpt redaction with the canonical secret-content patterns in `secrets.go` (adding a missing Google API key pattern in the process, and de-duplicating against an equivalent pattern that landed on `develop` independently) instead of a hand-copied subset, fixes a stale `docs/architecture.md` cross-reference, and renumbers the new ADR to 0019 after a collision with another ADR merged into `develop` in the meantime.

## Linked Issue

Closes #128

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- `TestScanForInstructionRisk*` (instructions_test.go) — prompt override, broad local read, exfiltration, silent destructive, credential collection, disable-safety escalation, oversized/symlinked content, excerpt redaction and truncation.
- `TestScanForInstructionRiskExcerptRedactsAdjacentSecret` / `TestScanForInstructionRiskExcerptRedactsGoogleAPIKey` — excerpt redaction covers the GitHub-token/password case and the previously-missing Google API key case.
- `TestScanForSecrets*` (secrets_test.go), including the Google API key case now shared with the redaction fix above.

```text
go build ./...
go vet ./...
go test ./...
```

All pass locally on the current head. Merged the latest `develop` into this branch to resolve conflicts (duplicate Google API key pattern/test, and an ADR-number collision — renumbered this PR's ADR to 0019).

Flagging separately from the code: no check runs are visible on this branch via the GitHub API historically — if `PR Policy`/`Test` still don't fire after this push, that's a workflow-trigger issue outside this diff's scope.

## Notes For Reviewers

- The redaction fix means the old hand-copied `secretLikeExcerptPatterns` list is gone; `redactSecretLike` now builds its pattern list from `secretContentPatterns` (secrets.go) plus one generic `key=value`/`key: value` pattern for credential names that aren't a distinct content shape.
- Re-requesting review from Rushilkr77 and priyam-jain-2002 on the latest head.


internal/packages/instructions.go
```diff
@@ -0,0 +1,356 @@
...
+		if loc == nil {
+			continue
+		}
+		seen[p.category] = true
+		findings = append(findings, InstructionFinding{
+			Path:     path,
+			Category: p.category,
+			Severity: baseSeverity(p.category),
+			Reason:   p.reason,
+			Excerpt:  excerptAround(content, loc[0], loc[1]),
```
priyam-jain-2002, line 309: This is useful review context, but printing the matched source line can accidentally leak secrets. For example, an exfiltration instruction may include a real token/password on the same line, and this excerpt is later printed by validate/inspect/enable/add. That breaks the existing safety posture in ScanForSecrets, which deliberately reports only path + reason and never echoes matched secret content. Please either omit excerpts for credential/exfiltration-style categories or run excerpts through the same secret redaction rules before storing/printing them, and add a regression test with a fake token on the flagged line.

docs/architecture.md
```diff
@@ -45,6 +45,8 @@ The receiver should be able to inspect these contents before enabling a package.
...
 
+Package instruction content — skills, workflows, agents, policies, adapters, and `setup.files[].template` — is separately scanned for risky agent-instruction patterns: prompt override, broad local reads, network/credential exfiltration, silent destructive operations, credential/session-data collection, and disable-safety language. Findings carry a hard-stop/warning severity split and are surfaced at validate, publish, inspect, and enable time; see [0016](decisions/0016-instruction-risk-scanning-severity-and-scope.md). Like the secret scan, this is a practical risk signal from a small documented pattern list, not a complete prompt-injection defense.
```
priyam-jain-2002, line 48: This new architecture paragraph says instruction-risk scanning is now part of the safety model, but docs/safety.md still has a dedicated "Instruction-Risk Scanning" section that says it is not implemented. Since this line points readers to docs/safety.md as the canonical safety model, please update that page in the same PR to describe the implemented scanner surfaces, warning vs block behavior, and the explicit limitations.

internal/cli/cli.go
```diff
@@ -867,7 +1088,7 @@ func runAdd(args []string, home string, stdin *bufio.Reader, stdout, stderr io.W
 	}
 
 	fmt.Fprintln(stdout)
-	enabled, err := enableRef(name, home, autoApprove, stdin, stdout, stderr)
+	enabled, err := enableRef(name, home, autoApprove, showsRiskInfo, stdin, stdout, stderr)
```
priyam-jain-2002, line 1098: In the risk-warning add flow this prints the warning/setup details before the single approval prompt, but enableRef still prints the same block again because preConfirmed skips only the read, not the output. That makes approved add runs show duplicate guidance. A clean fix would be to let preConfirmed also mean the assessment has already been displayed, or pass a small option into enableRef so it can apply setup without re-printing warnings/setup.

priyam-jain-2002 (changes requested):
Thanks for the thorough scanner work here. The implementation shape is mostly solid and the tests cover the main enable/add/validate paths well. I left three focused comments: please treat the excerpt leak as blocking before merge, then clean up the duplicated add output and update the canonical safety docs so the shipped behavior and documentation agree. Verified locally with go test ./... and go vet ./... .

priyam-jain-2002 (changes requested):
This is substantial security-sensitive behavior and is not merge-ready in its current state. Both required workflows are failing on the current head (`Test` and `PR Policy`), and the PR description is still the untouched template: no linked issue, package-impact classification, completed safety assertions, concrete verification results, or reviewer notes. Please first bring the branch up to date / resolve the failing checks, then complete the PR contract against #128 (including the exact test commands/results and safety implications). Once CI and policy are green, I’ll re-review the scanner semantics and enforcement paths rather than merging a security control on intent alone.

internal/packages/instructions.go
Raghavrajvanshi: Fixed in cf0a2e1. `Excerpt` is now run through a new `redactSecretLike` before it's ever stored — it reuses the same high-confidence patterns `ScanForSecrets` matches on (private key headers, AWS key IDs, GitHub token prefixes), plus a generic `key=value` / `key: value` form for credential-shaped names (token/secret/password/api_key/credential), so any plausible secret value adjacent to the flagged pattern gets replaced with `[REDACTED]` before printing. Updated the `InstructionFinding.Excerpt` doc comment to describe this instead of overclaiming safety.

Added `TestScanForInstructionRiskExcerptRedactsAdjacentSecret`, which flags a line carrying a fake GitHub token and a fake password and asserts neither value survives into the excerpt.

docs/architecture.md
Raghavrajvanshi: Fixed in cf0a2e1. Replaced `docs/safety.md`'s "Instruction-Risk Scanning: Not implemented in this build" section with an accurate description matching what's actually shipped: the six categories and their severities, that excerpts are redacted before printing, and where the scan runs in the lifecycle (blocks `validate`/`export`/`publish` via `BlockingCount()`, blocks/confirms at `enable`, surfaces in `inspect`) — mirroring ADR 0016. Also added "Instruction-risk scan" to the `lineage package validate` checklist alongside the existing "Secret scan" line, so it's not just described in prose but listed as one of the checks.

internal/cli/cli.go
Raghavrajvanshi: Fixed in cf0a2e1, via your first suggested option: extended what `preConfirmed` suppresses. Previously `preConfirmed` only skipped `enableRef`'s own `confirm()` read, not its `printEnableWarnings` call, so when `runAdd` had already printed the risk-warning/unscanned/setup-plan section and gotten its one merged yes, `enableRef` printed that same section again. Now `enableRef` skips the `printEnableWarnings` call entirely when `preConfirmed` is set — `runAdd` only ever sets it when it actually printed that section itself (`showsRiskInfo`), and a setup-only package (no warnings/unscanned) still gets its own separate setup-plan prompt from `enableRef`, unchanged from before.

Added `TestAddDoesNotDuplicateRiskWarningOutput`, which runs a package with both a risk warning and setup work through `add` and asserts each output section ("contains instructions flagged as risky", "wants to set up", "create file tasks.csv") appears exactly once, not twice.

priyam-jain-2002 (changes requested):
Re-reviewed the latest head (`cf0a2e1`) against the three previous findings and #128's safety boundary. The docs sync and duplicate `add` output fixes look directionally correct, and the excerpt regression test addresses the GitHub-token/password example. There is still one security blocker plus the PR-contract blocker:

1. `redactSecretLike` does **not** actually reuse all of `ScanForSecrets`' high-confidence patterns. `secrets.go` treats Google API keys (`AIza[0-9A-Za-z_-]{35}`) as a secret, but `secretLikeExcerptPatterns` omits that pattern. A risky instruction line containing a real Google API key can therefore be emitted verbatim in validate/inspect/enable/add output. That violates the safety property this fix is intended to restore and the new comments/docs currently overclaim ("same high-confidence patterns" / "always safe to print"). Please make the redaction source share/reuse the canonical secret-content patterns rather than copying a partial list, and add a regression case for an `AIza...` key so future additions to the secret scanner cannot silently diverge from excerpt redaction.

2. The PR description is still the untouched template. It still has no linked issue filled in, no package-impact classification, unchecked safety assertions, and no concrete verification/results. Please complete it against #128 before merge. I also cannot find any Actions/check run attached to this latest head, so merge readiness cannot be established from CI yet.

One smaller documentation cleanup: `docs/architecture.md` now says scanning is implemented, but the final sentence still says `docs/safety.md` documents what is not implemented yet and explicitly lists "instruction-risk scanning". Remove that stale phrase so the canonical docs do not contradict the paragraph immediately above it.

Keeping this blocked until the redaction invariant and PR contract are corrected. After that, this should get one final focused pass over the enforcement paths rather than another broad rewrite.

Rushilkr77 (changes requested):
Checked head `cf0a2e1` (still current, no commits since the last review) against the two blockers above — neither is resolved:

1. **Redaction gap still present.** `internal/packages/instructions.go`'s `secretLikeExcerptPatterns` has no Google API key pattern (`AIza[0-9A-Za-z_-]{35}`) — I grepped the whole tree at this head and `AIza` appears nowhere. `secrets.go`'s `ScanForSecrets` still treats that pattern as high-confidence, so a risky-instruction excerpt containing a real Google API key is still printed verbatim in `validate`/`inspect`/`enable`/`add` output. This is the core safety property #128 exists for; please make excerpt redaction share the canonical secret patterns (not a hand-copied subset) so this class of gap can't recur, and add the `AIza...` regression.
2. **PR description is still the unfilled template** — no linked issue, no package-impact/safety checkboxes, no verification section. Please complete it before this can be evaluated for merge readiness, and confirm CI is actually running on this head (still no check runs visible).

Keeping this blocked on both. Once the redaction source is unified and the PR contract is filled in, happy to do the final enforcement-path pass.

Raghavrajvanshi:
@Rushilkr77 Fixed in 0de500c — both blockers:

1. **Redaction gap.** Added the missing Google API key pattern (`AIza[0-9A-Za-z_-]{35}`) to `secretContentPatterns` in `internal/packages/secrets.go` — the canonical high-confidence secret list `ScanForSecrets` matches on. More importantly, per your "share the canonical patterns, not a hand-copied subset" suggestion: `redactSecretLike` in `internal/packages/instructions.go` no longer maintains its own copy of the pattern list. It now builds from `secretContentPatterns` directly plus one generic `key=value`/`key: value` pattern for credential names that aren't a distinct content shape. So a pattern added to `secretContentPatterns` in the future (a new provider's key format, etc.) can't silently miss excerpt redaction the way this one did. Added `TestScanForSecretsFlagsGoogleAPIKey` and `TestScanForInstructionRiskExcerptRedactsGoogleAPIKey` as regressions for the exact gap you found.
2. **PR description.** Filled in the template — linked to #128, package-impact/safety boxes checked, verification section filled in with the test commands and results.

On CI: confirmed via the GitHub API that there are 0 check runs registered against any commit on this branch, not just this head — looks like a workflow-trigger/repo-settings issue rather than something in this diff. Flagged it in the PR description; will look into it separately so it doesn't block this review further.

`go build ./...`, `go vet ./...`, and `go test ./...` all pass locally on `0de500c`.

Raghavrajvanshi:
@priyam-jain-2002 Addressed in 0de500c / be791a3:

1. **Redaction not actually sharing the canonical patterns** — fixed. `redactSecretLike` in `internal/packages/instructions.go` no longer keeps its own copy of the pattern list; it now builds from `secretContentPatterns` in `secrets.go` directly (plus one generic `key=value`/`key: value` pattern for credential names that aren't a distinct content shape). Also added the Google API key pattern (`AIza[0-9A-Za-z_-]{35}`) you called out to `secretContentPatterns` itself, and a regression test (`TestScanForInstructionRiskExcerptRedactsGoogleAPIKey`) for exactly that case. A pattern added to the canonical list going forward can't silently diverge from excerpt redaction again.
2. **PR description** — filled in against #128: linked issue, package-impact/safety boxes, verification section with commands run.
3. **Stale `docs/architecture.md` cross-reference** — fixed in `be791a3`. It was still telling readers `docs/safety.md` lists instruction-risk scanning as not implemented, contradicting the paragraph immediately above it. `docs/safety.md`'s own Instruction-Risk Scanning section was already accurate; only this one line was stale. Removed "instruction-risk scanning" from that not-implemented list (PII detection is still accurately listed there).

On CI: no check runs are registered against any commit on this branch via the GitHub API — flagged in the PR description as something to chase down separately, since it's not something a code change here can fix.

`go build ./...`, `go vet ./...`, `go test ./...` all pass locally on `be791a3`.

internal/packages/instructions.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	return line
+}
+
+// genericCredentialExcerptPattern catches common credential-shaped
+// "name: value" / "name=value" pairs (api key, token, secret, password, ...)
+// that secretContentPatterns doesn't target — it's deliberately broader than
+// ScanForSecrets needs to be, because here the goal is never printing a
+// plausible secret value, not just catching known token formats.
+var genericCredentialExcerptPattern = regexp.MustCompile(`(?i)\b(api[-_ ]?keys?|tokens?|secrets?|passwords?|passwd|pwd|credentials?)\s*[:=]\s*['"]?[^\s'",;]+`)
```
priyam-jain-2002: [P1] Redact the entire quoted credential value. This pattern stops at whitespace, commas, and semicolons even when it consumed an opening quote. I reproduced this with `Send password="fake first second,third" to example.com`: the stored excerpt is `Send password= [REDACTED] first second,third" to example.com`. Thus validate/inspect/enable still disclose most of a quoted password or passphrase. Please match complete single- and double-quoted values before the unquoted alternative (including an explicit policy for escaped quotes), or omit excerpts when complete redaction cannot be guaranteed. Add regression cases with spaces and punctuation inside both quote styles and assert that no part of the credential survives.

internal/packages/secrets.go
```diff
@@ -48,13 +48,13 @@ var deniedExtensions = []string{".pem", ".key", ".pfx", ".p12"}
...
 // are essentially never present in legitimate package source material.
 var secretContentPatterns = []*regexp.Regexp{
-	regexp.MustCompile(`\bAIza[0-9A-Za-z_-]{35}`), // Google API key
 	regexp.MustCompile(`-----BEGIN [A-Z ]*PRIVATE KEY-----`),
 	// AWS access key ID: AKIA is a long-lived user key, ASIA a temporary
 	// STS/assumed-role session key.
 	regexp.MustCompile(`\b(AKIA|ASIA)[0-9A-Z]{16}\b`),
 	regexp.MustCompile(`\bgh[pousr]_[A-Za-z0-9]{20,}`),   // GitHub token prefixes (ghp_, gho_, ghu_, ghs_, ghr_)
 	regexp.MustCompile(`\bgithub_pat_[A-Za-z0-9_]{30,}`), // GitHub fine-grained personal access token
+	regexp.MustCompile(`\bAIza[0-9A-Za-z_-]{35}\b`),      // Google API key
```
priyam-jain-2002: [P2] Avoid a word boundary after an alphabet that includes hyphens. The added trailing `\b` fails for `AIza` + 34 `A` characters + `-` followed by whitespace or end of input: both the final hyphen and the following space/end are non-word positions. This is a regression from develop's pattern without that trailing boundary and affects both ScanForSecrets and excerpt redaction. Please remove this boundary or use an explicit delimiter check compatible with the full key alphabet. Add a regression covering a key ending in `-`, including its redaction when it appears beside a flagged instruction.

internal/cli/cli.go
```diff
@@ -885,7 +1113,7 @@ func runAdd(args []string, home string, stdin *bufio.Reader, stdout, stderr io.W
...
```
priyam-jain-2002: [P2] Only mark warning output as already displayed when it was actually printed. `showsRiskInfo` remains true under `add --yes`, but the printing at lines 1104-1106 is inside `!autoApprove`. Passing it as `preConfirmed` then suppresses enableRef's printing too. I reproduced a successful `add <local archive> --yes` for a prompt-override package with no risk warning anywhere in the output; unscanned-file notices and accompanying setup previews are lost the same way. Please pass `showsRiskInfo && !autoApprove` here, or print the assessment for both modes and reserve --yes for skipping the read. Add --yes coverage asserting warnings/unscanned notices are shown exactly once, consistent with enable --yes.

priyam-jain-2002 (changes requested):
Reviewed the latest head 05090d7 line by line, including the fixes to the earlier comments. The interactive duplicate-output issue and stale safety documentation are addressed, and sharing the canonical secret patterns is a good improvement. Three remaining issues are detailed inline: quoted credential values are only partly redacted; Google keys ending in a hyphen evade the changed pattern; and add --yes suppresses all risk/unscanned warnings. The unchanged head passes go test ./... and go vet ./.... I added temporary targeted reproductions in an isolated checkout; each of the three cases fails as described. Please add regression coverage for these cases with the fixes.

Raghavrajvanshi:
PR description updated to match the current template — it was missing the `## Merge Method` section, which is why `PR Policy` failed on the last push (confirmed from the job log: `PR must include a Merge Method section.`). That's fixed now.

Separately, and this needs a maintainer: the `PR Policy` workflow run triggered by that edit is stuck at `action_required` (https://github.com/agentic-lineage/lineage/actions/runs/35339625813), not failing or running. This PR is from a fork (`Raghavrajvanshi/lineage`), and this repo's Actions settings require a maintainer with admin rights to manually approve each workflow run from a fork PR — something I don't have permission to do (confirmed via the API: `403 Must have admin rights to Repository`).

This is almost certainly the actual explanation for "no check runs visible on this branch" that both reviewers and I flagged earlier — every push from this fork needs a manual approval click before any check (`Test`, `PR Policy`) even starts.

Could someone with admin access either:
- Approve the pending run (Actions tab → this run → "Approve and run"), or
- Enable auto-run for fork PR workflows (Settings → Actions → General → "Fork pull request workflows")

so CI actually reflects the current head?

Raghavrajvanshi:
@priyam-jain-2002 Thanks for the reproductions — all three fixed in 08c71cc:

1. **Quoted credential truncation.** `genericCredentialExcerptPattern` now tries a complete double-quoted value, then a complete single-quoted value, before falling back to the unquoted run — in that order — so `password="fake first second,third"` gets the whole quoted value redacted instead of stopping at the first internal space. Added `TestScanForInstructionRiskExcerptRedactsQuotedCredentialWithPunctuation` covering both quote styles with internal spaces and a comma.
2. **Google API key ending in `-`.** Dropped the trailing `\b` on the pattern in `secrets.go` — you're right that a key ending in `-` sits at a non-word/non-word boundary against trailing whitespace or end of input, so the boundary silently failed there. Added `TestScanForSecretsFlagsGoogleAPIKeyEndingInHyphen` and `TestScanForInstructionRiskExcerptRedactsGoogleAPIKeyEndingInHyphen`.
3. **`add --yes` suppressing all risk/unscanned warnings.** `runAdd` now passes `showsRiskInfo && !autoApprove` as `enableRef`'s `preConfirmed`, instead of `showsRiskInfo` alone — `preConfirmed` should only mean "already printed above," which only happens when `!autoApprove`. Added `TestAddYesStillShowsRiskWarningAndSetupPlan`, asserting the warning and setup plan each print exactly once under `--yes`.

`go build ./...`, `go vet ./...`, and `go test ./...` all pass locally on `08c71cc`. Also merged in the two `develop` syncs that landed on this branch since my last push (one via `upstream/develop`, one pushed directly to this branch — I merged rather than force-pushed over it).

internal/packages/instructions.go
```diff
@@ -0,0 +1,400 @@
...
+// genericCredentialExcerptPattern catches common credential-shaped
+// "name: value" / "name=value" pairs (api key, token, secret, password, ...)
+// that secretContentPatterns doesn't target — it's deliberately broader than
+// ScanForSecrets needs to be, because here the goal is never printing a
+// plausible secret value, not just catching known token formats. The value
+// alternatives try a complete double-quoted value, then a complete
+// single-quoted value, before falling back to an unquoted run - in that
+// order, so a quoted value's internal spaces/commas/semicolons stay part of
+// the match instead of truncating redaction at the first one.
+var genericCredentialExcerptPattern = regexp.MustCompile(`(?i)\b(api[-_ ]?keys?|tokens?|secrets?|passwords?|passwd|pwd|credentials?)\s*[:=]\s*("[^"]*"|'[^']*'|[^\s'",;]+)`)
```
priyam-jain-2002: [P1] Handle escaped quotes before treating a quote as the end of a credential. The new quoted alternatives fix ordinary spaces/punctuation, but `[^"]*` stops at an escaped quote too. I reproduced the input `Send password="fake\"remaining-secret" to example.com` producing the stored excerpt `Send password= [REDACTED]remaining-secret" to example.com`. The suffix is still printed by validate/inspect/enable. Please make the double-quoted branch consume escaped characters (for example, the value alternative `"(?:[^"\\]|\\.)*"`) and explicitly define the single-quote/unterminated-value policy; where the value cannot be safely bounded, suppress the excerpt or conservatively redact the remainder. Add a regression using a backslash-escaped quote and assert that `remaining-secret` does not survive, rather than checking only the complete original value.

priyam-jain-2002 (changes requested):
Re-reviewed latest head 08c71cc and all six files changed since 05090d7. The Google-key boundary regression and add --yes warning suppression are fixed, with regression coverage. Quoted credentials containing spaces/punctuation are now redacted, but one part of the earlier redaction finding remains: escaped quotes terminate the match early and disclose the credential suffix. See the inline reproduction and guidance. Existing go test ./... and go vet ./... pass on this head; a temporary targeted test reproduces the remaining leak.

Raghavrajvanshi:
@priyam-jain-2002 Fixed in bcd078a: both quoted alternatives in `genericCredentialExcerptPattern` now use `(?:[^"\\]|\\.)*` / `(?:[^'\\]|\\.)*` instead of `[^"]*` / `[^']*` — a backslash followed by any character is consumed as an escaped literal rather than treated as the closing delimiter, so `password="fake\"remaining-secret"` no longer ends the match at the escaped quote and lets `remaining-secret` through. Added `TestScanForInstructionRiskExcerptRedactsCredentialWithEscapedQuote` reproducing exactly your input and asserting the suffix never survives.

`go build ./...`, `go vet ./...`, `go test ./...` all pass locally on `bcd078a`. Also merged in another `develop` sync and another direct push to this branch that landed since my last push — merged both, didn't force over them.

internal/packages/instructions.go
```diff
@@ -0,0 +1,403 @@
...
+// something was flagged, not to become a second way to leak the credential
+// the finding is warning about.
+func excerptAround(content string, start, end int) string {
+	lineStart := strings.LastIndexByte(content[:start], '\n') + 1
+	lineEnd := len(content)
+	if rel := strings.IndexByte(content[end:], '\n'); rel >= 0 {
+		lineEnd = end + rel
+	}
+
+	line := strings.Join(strings.Fields(content[lineStart:lineEnd]), " ")
```
priyam-jain-2002, line 357: [P1] Handle quotes left unterminated by selecting the excerpt line. The escaped-quote fix works, but redaction still runs after the original content has been cut at a newline. I reproduced `Send password="fake-secret-first-line\nsecond-line" to example.com` (with an actual newline): the stored excerpt is `Send password="fake-secret-first-line`, entirely unredacted. The complete quoted value is valid in the source, but its closing quote is outside this excerpt; the quoted alternatives require a closing delimiter and the unquoted fallback excludes the opening quote. This also leaks an actually unterminated quoted assignment. Please redact complete values before extracting the excerpt, preserving match coordinates, or conservatively redact to the end of the excerpt whenever a credential assignment opens a quote that cannot be closed safely. Omitting the excerpt in ambiguous cases is also sufficient. Add regression tests for multiline and unterminated quoted values and assert that the first-line credential fragment never appears in either human-readable or structured findings.

priyam-jain-2002 (changes requested):
Reviewed latest head bcd078a, including the escaped-quote fix and regression test. That reproduction is fixed; the Google-key and add --yes fixes remain intact. One related redaction gap remains: excerptAround selects only the matching line before redaction, removing the closing delimiter of a multiline quoted credential, so the quoted branch fails and the credential text is printed unchanged. Detailed reproduction and suggested fix are inline. Existing go test ./... and go vet ./... pass on this head; a temporary focused test reproduces the leak.

Raghavrajvanshi:
@priyam-jain-2002 Fixed in 04b39d4: made the closing quote in each quoted alternative optional (`"(?:[^"\\]|\\.)*"?` / `'(?:[^'\\]|\\.)*'?`). The greedy body still stops at a real closing quote when the excerpt line contains one (existing behavior unchanged), but when the excerpt's single-line bound cuts the value off before its closing quote — or the assignment is genuinely unterminated — it now consumes to the end of the available text instead of failing to match at all, so the fragment gets redacted conservatively rather than printed in full.

Added `TestScanForInstructionRiskExcerptRedactsMultilineQuotedCredential` (reproducing your exact case: opening quote on the flagged line, closing quote on the next) and `TestScanForInstructionRiskExcerptRedactsUnterminatedQuotedCredential` (no closing quote anywhere in the file).

`go build ./...`, `go vet ./...`, `go test ./...` all pass locally on `04b39d4`.

priyam-jain-2002 (approved):
Re-reviewed latest head 04b39d4, including both changed files line by line. The remaining multiline/unterminated quoted-credential reproduction is addressed: redaction now conservatively consumes quoted values without requiring a closing delimiter in the excerpt, and both cases have regression tests. The prior escaped-quote, Google-key boundary, and add --yes findings remain addressed. No further actionable findings in this review. Verified locally with go test ./... and go vet ./...; both pass. GitHub reports no checks on this branch, so this approval reflects code review and local verification, not a passing CI run.

Rushilkr77 (approved):

PR #229, opened 2026-08-31
secrets: flag Google API keys (AIza prefix)

### LINKED ISSUE #230: Secret scanner misses Google API keys (AIza prefix)
## Gap

`ScanForSecrets` in `internal/packages/secrets.go` currently flags private-key headers, AWS access key IDs (`AKIA`/`ASIA`), and GitHub tokens (`ghp_`/`github_pat_`), but not Google API keys.

Google API keys are 39-char tokens shaped `AIza` followed by 35 chars of `[A-Za-z0-9_-]`, issued from the Google Cloud Console. They are essentially never present in legitimate package source material — the same bar the existing content patterns set under ADR 0009.

## Fix direction

Add `regexp.MustCompile(`\bAIza[0-9A-Za-z_-]{35}`)` to `secretContentPatterns`, with a test that mirrors the AWS key ID fixtures: split the literal (`"AIza" + strings.Repeat("A", 35)`) so GitHub push protection does not flag the test source itself.

## Confidence

High on the gap. This is an extension of the existing documented list, not a new scanning engine, so it stays within the "documented list, not exhaustive" tradeoff of ADR 0009.

MsfPablo:
## Summary

`ScanForSecrets` flags private-key headers, AWS access key IDs, and GitHub tokens, but not Google API keys. This adds the Google API key content pattern (`AIza` + 35 chars of `[A-Za-z0-9_-]`) to `secretContentPatterns`, with a test mirroring the existing AWS key ID fixtures.

Closes #230.

## Package Impact

## Safety

Findings still carry only the path and reason, never the matched value — `TestScanForSecretsNeverIncludesMatchedValue` continues to cover that.

## Verification

Relevant test cases:

- `TestScanForSecretsFlagsGoogleAPIKey` (new) — a fake `AIza`-prefixed key in `references/config.txt` produces a finding.
- `TestScanForSecretsFlagsAWSKeyID` — unchanged, `AKIA` still matches.
- `TestScanForSecretsIgnoresSafeContent` — unchanged, no new false positives.

```text
go build ./...
go test ./internal/packages/
go test ./...
ok  	github.com/agentic-lineage/lineage/internal/packages	0.612s
```

## Notes For Reviewers

Pattern placed first in the list (before the PEM header) to avoid the anchor develop already used for the AWS and GitHub fine-grained token additions. The new test fixture splits the literal (`"AIza" + strings.Repeat("A", 35)`) the same way the existing `AKIA` fixtures do, to keep push protection from flagging the test file itself. A 3-way merge against `upstream/develop` is clean.


priyam-jain-2002 (approved):
Reviewed the scanner change end to end. The Google API key pattern is high-confidence, the test covers the new detection path without embedding a contiguous key-shaped literal in source, and existing behavior still avoids returning matched secret values. I left one non-blocking inline note to keep the secret-scanning ADR/documented list in sync with this new pattern. Validation run locally: `go test ./internal/packages`, `go test ./...`, `go build ./...`, `go vet ./...`, and `git diff --check origin/develop...HEAD`.

internal/packages/secrets.go
```diff
@@ -48,6 +48,7 @@ var deniedExtensions = []string{".pem", ".key", ".pfx", ".p12"}
 // secretContentPatterns are high-confidence content signatures: things that
 // are essentially never present in legitimate package source material.
 var secretContentPatterns = []*regexp.Regexp{
+	regexp.MustCompile(`\bAIza[0-9A-Za-z_-]{35}`), // Google API key
```
priyam-jain-2002, line 51: This adds Google API keys to the actual scanner list, but `docs/decisions/0009-secret-scanning-is-a-documented-list-not-an-engine.md` still describes the documented content patterns as private key headers, AWS access key IDs, and GitHub token prefixes only. Since that ADR explicitly frames this as a small documented list, it would be worth updating the design note so future maintainers do not treat Google API keys as an undocumented exception.

priyam-jain-2002: Adding them in a larger documentation update pr.

priyam-jain-2002:
Thanks for the contribution, @MsfPablo.

This is a clean, focused extension of the existing secret-scanning model: Google API keys now get the same high-confidence treatment as the existing AWS and GitHub credential patterns, with a regression test that avoids embedding a contiguous key-shaped fixture in the repository.

I’ve reviewed the change end to end and verified the package tests, full test suite, build, vet, and diff checks. The ADR documentation sync is being handled separately as part of the broader documentation update, so it does not need to block this change.

Merging this in. Thanks for identifying the gap and for the careful implementation.

PR #254, opened 2026-09-01
packages: reject duplicate skill exports across enabled packages

### LINKED ISSUE #246: Detect duplicate skill exports across enabled packages
## Problem

The current dependency model identifies required skills by bare name only. On `develop`, `Requires.Skills` is `[]string`, and `ValidateDependencies` builds a boolean set of skill names across enabled packages. That answers only "does some enabled package provide this name?".

It does not establish a unique provider. If two enabled packages both export a skill named `research`, dependency validation succeeds even though downstream discovery/materialization can encounter two competing definitions. A dependency on `research` is therefore ambiguous and can be satisfied by whichever definition wins later resolution order.

This issue intentionally addresses the bounded correctness problem — **duplicate export detection** — without turning it into a full package-manager resolver.

## Proposed behavior

When validating/resolving the complete enabled package set, build a map from exported skill name → providing package(s). If more than one enabled package provides the same skill name, fail before materialization/launch with an actionable conflict error that names the skill and all conflicting packages.

Example:

```text
cannot enable package set: skill "research" is exported by both "foo" and "bar"
```

Do not silently choose first/last provider. Do not make filesystem/discovery iteration order part of dependency semantics.

A future namespace/qualified-reference design may provide an explicit disambiguation mechanism; until one exists, ambiguity should be rejected.

## Area

Package discovery

## Compatibility notes

A valid enabled set with unique skill names behaves exactly as today. This can make previously ambiguous package combinations fail explicitly; that is the intended safety/correctness change.

## Out of scope

`Requires.Skills` also has no version constraints, so a package cannot express `research >=1.2,<2`. That is a separate, substantially larger dependency-resolution problem. Do not add semver range solving as part of this issue. Track/design it only when Lineage has a concrete compatibility requirement and a resolved dependency-state model to build on.

## Acceptance criteria

- [ ] Validation detects when 2+ enabled packages export the same skill name.
- [ ] The error identifies the conflicting skill and provider package names.
- [ ] Conflict detection is deterministic and independent of package/discovery iteration order.
- [ ] A package does not conflict with itself merely because it requires a skill it exports.
- [ ] Unique-provider dependency behavior remains unchanged.
- [ ] Unit tests cover two-provider conflict, 3+ provider conflict, no conflict, and required-skill resolution in the presence of unique providers.
- [ ] Documentation/error guidance makes clear that version-range resolution is not implemented by this change.

## Implementation guidance

`internal/packages/dependencies.go` is the natural enforcement boundary because it already validates the full package set. Keep the check there (or in a single resolver it delegates to) rather than adding separate duplicate checks to every provider/materialization path.

MsfPablo:
## Summary

- Closes #246. `ValidateDependencies` built a `map[string]bool` over exported skill names, so it could only answer "does *some* enabled package provide this name?". Two enabled packages exporting the same skill name passed validation, leaving downstream discovery to pick a definition by iteration order — exactly the ambiguity #246 calls out.
- Tracks the providing package names per skill and fails when a name has more than one provider: `cannot enable package set: skill "research" is exported by both "foo" and "bar"`. The error names the skill and every conflicting package, and conflicts are reported deterministically (sorted by skill name, then provider name) so the message is stable regardless of package or skill input order.

Behavior preserved:
- A package requiring a skill it also exports still validates (single-provider case, not a duplicate).
- Required-skill resolution is unchanged: a skill is available iff at least one enabled package exports it; the missing-skill error message is untouched.

Scope is bounded to duplicate-export detection. Matching is by bare name only; version-range resolution is not implemented, so two packages exporting the same name at different versions still conflict. A future namespace/qualified-reference design can introduce explicit disambiguation; until then, ambiguity is rejected rather than silently resolved.

`ValidateDependencies` is the single enforcement point — it runs from `runtime/plan.go` (before launch) and `cli.go` (enable/validate paths), so the check fires before materialization in every entry that already validated dependencies.

## Linked Issue

Closes #246

- [ ] The linked issue is assigned to me.

## Package Impact

- [ ] Documentation only
- [x] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- `TestValidateDependenciesRejectsDuplicateSkillExport`: two packages exporting the same skill fail, and the error names the skill and both packages.
- `TestValidateDependenciesDuplicateExportIsOrderIndependent`: swapping package input order produces the same error.
- `TestValidateDependenciesRejectsThreeOrMoreProviders`: three providers of one skill all appear in the error.
- `TestValidateDependenciesSelfProvidesAndRequiresIsNotADuplicate`: a package requiring a skill it also exports is the satisfied-by-own-skill case, not a duplicate.
- `TestValidateDependenciesUniqueProvidersAndResolvedRequirementsPass`: a unique provider plus a satisfied cross-package requirement still passes.
- Existing tests (`SatisfiedByOwnSkill`, `SatisfiedByAnotherPackage`, `MissingSkillFails`) unchanged and still pass.

```text
cd internal/packages && go test ./...
go test ./...
go vet ./...
```

- Automated tests apply; the cases above cover the new duplicate-export detection and confirm prior dependency-validation behavior is preserved.

## Notes For Reviewers

- The deterministic sorting (skill name, then provider name) is deliberate so CI/failure messages do not depend on filesystem or map iteration order.
- Version-range disambiguation is intentionally out of scope; this PR only removes the silent order-dependent pick that #246 describes.

### DIFF
--- a/internal/packages/dependencies.go
+++ b/internal/packages/dependencies.go
@@ -1,26 +1,81 @@
 package packages
 
-import "fmt"
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
 
-// ValidateDependencies checks that every package's Requires.Skills is
-// satisfied somewhere across pkgs — its own skills, or another package's.
-// It's meant to run against the full set of packages that will be enabled
-// together (not one package in isolation), since a required skill can
-// legitimately be provided by a different enabled package.
+// ValidateDependencies checks two invariants over the set of packages that will
+// be enabled together:
+//
+//  1. No two enabled packages export the same skill name. A skill name must
+//     resolve to exactly one providing package, otherwise the launcher cannot
+//     decide which implementation to wire into the run — and a silent,
+//     order-dependent pick is worse than failing loudly. On conflict it
+//     returns an error naming the skill and every package that exports it.
+//
+//  2. Every package's Requires.Skills is satisfied somewhere across pkgs — its
+//     own skills, or another package's. It runs against the full enabled set
+//     (not one package in isolation) because a required skill can legitimately
+//     come from a different enabled package.
+//
+// Skill matching is by bare name only. Version-range resolution is not
+// implemented: two packages exporting the same name at different versions still
+// conflict, since there is no rule to pick between them.
 func ValidateDependencies(pkgs []Package) error {
-	available := map[string]bool{}
+	// skill -> package names that export it, in input order (sorted for the
+	// error message below). A skill with more than one provider is a conflict.
+	providers := map[string][]string{}
 	for _, pkg := range pkgs {
 		for _, skill := range pkg.Skills {
-			available[skill] = true
+			providers[skill] = append(providers[skill], pkg.Manifest.Name)
 		}
 	}
 
 	for _, pkg := range pkgs {
+		// A package does not conflict with itself merely because it requires a
+		// skill it also exports; that's the satisfied-by-own-skill case.
 		for _, required := range pkg.Manifest.Requires.Skills {
-			if !available[required] {
+			if _, ok := providers[required]; !ok {
 				return fmt.Errorf("package %q requires skill %q, which is not provided by any enabled package", pkg.Manifest.Name, required)
 			}
 		}
 	}
+
+	// Report duplicate exports deterministically: sort by skill name so the
+	// error is stable regardless of package/skill input order.
+	var conflicts []string
+	for skill, names := range providers {
+		if len(names) < 2 {
+			continue
+		}
+		sort.Strings(names)
+		conflicts = append(conflicts, fmt.Sprintf("skill %q is exported by %s", skill, providerPhrase(names)))
+	}
+	if len(conflicts) > 0 {
+		sort.Strings(conflicts)
+		return fmt.Errorf("cannot enable package set: %s", strings.Join(conflicts, "; "))
+	}
 	return nil
+}
+
+// providerPhrase renders a list of package names as a quoted phrase suitable
+// for a conflict error: ["foo","bar"] -> `both "foo" and "bar"`, and
+// ["a","b","c"] -> `"a", "b", and "c"`.
+func providerPhrase(names []string) string {
+	quoted := make([]string, len(names))
+	for i, n := range names {
+		quoted[i] = fmt.Sprintf("%q", n)
+	}
+	switch len(quoted) {
+	case 0:
+		return ""
+	case 1:
+		return quoted[0]
+	case 2:
+		return "both " + quoted[0] + " and " + quoted[1]
+	default:
+		return strings.Join(quoted[:len(quoted)-1], ", ") + ", and " + quoted[len(quoted)-1]
+	}
 }
--- a/internal/packages/dependencies.go
+++ b/internal/packages/dependencies.go
@@ -23,13 +23,23 @@ import (
 // Skill matching is by bare name only. Version-range resolution is not
 // implemented: two packages exporting the same name at different versions still
 // conflict, since there is no rule to pick between them.
+//
+// The same package may be supplied more than once when two enabled refs resolve
+// to the same content; dedupePackages collapses those into a single provider so
+// the duplicate is not mistaken for a self-conflict. When two genuinely
+// different packages share a manifest name, the conflict error widens the
+// identity past the bare name — to name@version, and to name@version with the
+// resolved path when name@version still collides — so the operator can tell
+// which copy to disable instead of reading `both "foo" and "foo"`.
 func ValidateDependencies(pkgs []Package) error {
-	// skill -> package names that export it, in input order (sorted for the
-	// error message below). A skill with more than one provider is a conflict.
-	providers := map[string][]string{}
+	pkgs = dedupePackages(pkgs)
+
+	// skill -> packages that export it, in input order. A skill with more than
+	// one provider is a conflict.
+	providers := map[string][]Package{}
 	for _, pkg := range pkgs {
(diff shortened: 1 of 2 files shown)


internal/packages/dependencies.go
```diff
@@ -1,26 +1,81 @@
...
+// conflict, since there is no rule to pick between them.
 func ValidateDependencies(pkgs []Package) error {
-	available := map[string]bool{}
+	// skill -> package names that export it, in input order (sorted for the
+	// error message below). A skill with more than one provider is a conflict.
+	providers := map[string][]string{}
 	for _, pkg := range pkgs {
 		for _, skill := range pkg.Skills {
-			available[skill] = true
+			providers[skill] = append(providers[skill], pkg.Manifest.Name)
```
priyam-jain-2002: This records only `pkg.Manifest.Name` as the provider identity. If two enabled refs resolve to different package copies or versions with the same manifest name, the conflict error becomes something like `skill "research" is exported by both "foo" and "foo"`, which does not identify the conflicting packages and is hard to act on. Consider tracking a display identity such as `name@version` plus path when needed, and optionally deduplicating the exact same resolved package before counting providers.

priyam-jain-2002 (changes requested):
Thanks for the focused fix here. The main duplicate-export behavior matches #246 and the test suite is green locally, but I found one actionable edge case that should be fixed before merge.

MsfPablo:
Good catch — pushed a fix in 1eeee97.

Provider identity now widens only as far as it needs to: bare name when the names are unique, `name@version` when they collide, and `name@version (path)` when both still match. So the ambiguous `both "foo" and "foo"` case is gone.

I also added the dedup you suggested: two enabled refs that resolve to the same package (matching content digest, falling back to the on-disk path) collapse into one provider, so that doesn't get counted as a self-conflict. Tests cover both the version-differs and path-differs cases plus the dedup paths.

internal/packages/dependencies.go
```diff
@@ -1,26 +1,162 @@
...
+// pointing at one package are counted as a single provider, not a
+// self-conflict. Two packages are the same when their content digests match
+// (the digest covers the manifest and every content file, so identical content
+// always matches); when a digest is not available, the on-disk path is used
+// instead. The first occurrence wins; later duplicates are dropped.
+func dedupePackages(pkgs []Package) []Package {
+	seen := make(map[string]struct{}, len(pkgs))
+	out := make([]Package, 0, len(pkgs))
+	for _, pkg := range pkgs {
+		key := pkg.Digest
```
priyam-jain-2002, line 84: Deduping by `pkg.Digest` collapses two different on-disk package copies that happen to have identical content, even though the rest of the run still receives both original `Package` entries. That means `ValidateDependencies` can now pass, but `materialize.desiredSkillDirs` later sees the same staged directory backed by two different source paths and returns a collision error before launch. In practice, `lineage package add` can save an enabled set that validation accepted but `lineage run` cannot materialize. Either limit this dedupe to the same resolved path, or make the deduped package set flow through to the plan/materialization paths as well.

priyam-jain-2002 (changes requested):
Thanks for the update. The previous same-name error message issue is addressed, and the local test/vet checks pass, but the new digest-based dedupe creates a validation/materialization mismatch that should be fixed before merge.

priyam-jain-2002 (approved):
Reviewed the dependency-validation change end to end. The fail-closed behavior is appropriate here: ambiguous bare skill exports are rejected before materialization rather than left to iteration order. The implementation also preserves existing required-skill semantics, makes conflict reporting deterministic, and handles duplicate refs to the same resolved package without turning them into false conflicts. The added tests cover two/three-provider conflicts, order independence, self-provisioning, same-content deduplication, and identity disambiguation. PR Policy and Test checks are green on this head. Approved for squash merge into `develop`.

priyam-jain-2002 (changes requested):
Correction after re-checking the current head against the unresolved inline thread: the earlier provider-identity issue is fixed, but the digest-based dedupe still creates the validation/materialization mismatch called out in the remaining thread. `ValidateDependencies` can collapse identical-content packages from different paths while downstream materialization still receives both original package entries. Please either dedupe only the same resolved path here, or propagate one canonical deduped package set through planning/materialization. Keeping this blocked until that invariant is consistent end to end.

Rushilkr77 (changes requested):
Re-checked the current head (`cf7a843`, just a develop merge, no new fix commits since 2026-09-05) against the unresolved inline thread on `internal/packages/dependencies.go:84`.

The identity-widening fix (name@version, then +path) is solid and covers the ambiguous `both "foo" and "foo"` case well — that part is good to go.

The digest-based dedupe issue from the 2026-09-05 correction is still present: `dedupePackages` collapses packages by `Digest`/`Path` before `ValidateDependencies` ever sees them, so two enabled refs resolving to identical content are silently merged for validation purposes. But nothing shows this collapsed view is what actually reaches materialization downstream — if the launch/materialize path still receives both original `Package` entries (as the original finding noted), you get a set that validates clean but then double-processes or double-writes a skill directory at apply time. Please either:
- thread the deduped slice through to whatever consumes `pkgs` after validation, so the same collapsed set is what's materialized, or
- confirm (with a test exercising the full validate→materialize path, not just `ValidateDependencies` in isolation) that this can't actually happen, and explain why in a comment near `dedupePackages`.

Also worth resolving in the open: #276 is a duplicate of this PR for #246, less mature (no identity-widening, no dedupe) and has been asked to close in favor of this one — flagging so this doesn't sit blocked on a stale duplicate.

Blocked on the dedupe/materialization consistency point; everything else here looks ready.

PR #271, opened 2026-09-08
feat(packages): declare and validate MCP dependencies

### LINKED ISSUE #245: Support MCP servers as a first-class package dependency kind
## Problem

Lineage packages can declare skills, agents, workflows, policies/references through the package model, but an MCP server is not currently a first-class dependency kind. A package whose intended behavior depends on an MCP server therefore has no portable, inspectable way to declare that requirement in `lineage.yaml`.

This matters because Lineage's package thesis is behavioral distribution, not just file copying: external tool/runtime dependencies that are necessary to reproduce a workflow should be declared before enablement rather than hidden in prose or provider-local configuration.

This is a parity/completeness gap, not a reason to create a second MCP-specific dependency system.

## Proposed behavior

Extend the manifest with an MCP dependency declaration, likely under a general dependency surface such as `dependencies.mcp` (final schema shape should follow existing manifest conventions).

A declaration should be able to express the server identity and the explicit connection mode needed by the package, for example:

- `stdio` with executable/arguments;
- `streamable-http` with endpoint metadata;
- `sse` where still supported/needed.

Do **not** put credentials/secrets into the package manifest. Authentication requirements should be declarative and receiver-provided through the existing/future setup and capability boundaries.

The design should decide how MCP dependencies participate in:

- manifest validation and inspect output;
- dependency resolution/materialization;
- provider adapters that can consume MCP configuration;
- resolved dependency state/snapshots;
- capability/safety review before enablement.

## Area

Package format

## Compatibility notes

MCP is shared package semantics, but materialization is provider-specific. Core should represent the dependency; Claude/Codex/Cursor/Cline/etc. adapters should translate only where the provider supports that transport/configuration. Unsupported provider mappings should fail or surface clearly rather than silently dropping the dependency.

The manifest should not depend on any one MCP discovery mechanism. Explicit server/transport/connection semantics keep the package contract stable while discovery conventions evolve.

## Explicit cross-references / boundaries

- Reuse #132 for capability enforcement rather than adding MCP-only permission logic.
- Reuse #134 for receiver-facing safety summaries.
- Reuse #135 for registry trust states.
- Capture MCP dependencies in the resolved dependency/materialization state designed by #200 rather than creating a second lockfile.

## Acceptance criteria

- [ ] `lineage.yaml` has a documented, validated MCP dependency representation.
- [ ] MCP dependency declarations never require embedding credentials in the package.
- [ ] `package validate` / inspect surfaces server identity, transport, and relevant connection metadata safely.
- [ ] Provider adapters explicitly report supported vs unsupported MCP materialization.
- [ ] MCP dependencies participate in the same capability/safety review path as other executable/tool dependencies.
- [ ] Resolved MCP state uses the dependency-state model from #200 rather than a parallel lockfile.
- [ ] Tests cover schema validation, unsupported transports/providers, secret non-leakage, and deterministic materialization where supported.
- [ ] Docs include at least one stdio and one remote-server example without real credentials.

priyam-jain-2002:
## Summary

Add a portable, inspectable MCP dependency contract to Lineage packages. The manifest can declare stdio, streamable-http, and sse servers; validation rejects unsafe/unmodelled configuration and exposes required network capability before enablement.

## Linked Issue

Closes #245

- [x] This PR is ready for review: the linked issue and test plan are complete.

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- Valid stdio and remote MCP declarations, including required declared network capability.
- Invalid transport shapes, embedded URL credentials, unsupported MCP fields, duplicate server names, and missing stdio commands.
- Inspect and structured-report visibility.
- Provider launch refusal when materialization is unsupported, preventing silent omission.

```text
git diff --check
go test ./...
```

The first command passed. `go test ./...` is required and will run in the GitHub Actions Linux/Windows matrix; the initial run exposed a stale-base error and this branch has now been rebuilt on current `develop`.

## Notes For Reviewers

- This deliberately defines the package contract and safety boundary, not native provider MCP configuration merging.
- Claude, Codex, and the other adapters fail explicitly when a package declares MCP dependencies they cannot materialize. A follow-up needs provider-specific merge/reconciliation state before materialization can be added safely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Lineage
 
-Package and share Claude Code, Codex, Windsurf, Aider, and Cline workflow environments.
+Package and share Claude Code and Codex workflow environments.
 
 A reusable agent workflow usually depends on more than one prompt or skill:
 skills, workflow steps, agents, policies, references, setup files, declared
@@ -32,15 +32,12 @@ Use Lineage when you want to:
   deterministic `.tgz` archive.
 - Let receivers inspect package contents and declared capabilities before
   enabling the workflow locally.
-- Run the packaged workflow through Claude Code, Codex, Windsurf, Aider, or Cline
-  without rebuilding the environment by hand in each project.
+- Run the packaged workflow through Claude Code or Codex without rebuilding the
+  environment by hand in each project.
 
 Lineage is provider-adjacent, not provider-owned. It prepares local files for
 the agent provider the receiver already uses; it does not replace Claude Code,
-Codex, Windsurf, Aider, Cline, or the user's agent account. Windsurf and Cline
-support is project-scoped and config/materialization-only: their `lineage run`
-and workflow forms stage project files and exit without resolving or launching
-their native GUIs.
+Codex, or the user's agent account.
 
 ## Install
 
@@ -79,9 +76,6 @@ lineage package validate ./resume-workflow
 lineage enable ./resume-workflow
 lineage run claude --dry-run
 lineage run codex --dry-run
-lineage run windsurf --dry-run
-lineage run aider --dry-run
-lineage run cline --dry-run
 ```
 
 Publish it to the Lineage registry:
@@ -183,6 +177,37 @@ package/
 These folders are intentionally plain. A receiver should be able to open the
 package and see what it contains before enabling it.
 
+### MCP dependencies
+
+Packages can declare an MCP server requirement under `dependencies.mcp`. The
+declaration is inspectable before enablement; it contains connection metadata,
+not credentials. Remote MCP hosts must also be declared in `capabilities.network`.
+
+```yaml
+dependencies:
+  mcp:
+    - name: github
+      transport: stdio
+      command: npx
+      args: ["-y", "@modelcontextprotocol/server-github"]
+      auth: receiver
+    - name: docs
+      transport: streamable-http
+      url: https://mcp.example.com/mcp
+      auth: receiver
+capabilities:
+  network: [mcp.example.com]
+```
+
+Supported transports are `stdio`, `streamable-http`, and `sse`. Lineage does
+not accept tokens, passwords, headers, or environment values in this package
+surface; receivers configure authentication locally.
+
+The current Claude and Codex adapters do not yet materialize MCP configuration.
+`lineage run` therefore stops with a clear error rather than silently omitting
+a declared server; native configuration merging belongs in the follow-up adapter
+work.
+
 ## How Lineage Fits Claude, Codex, And Other Agents
 
 Lineage keeps provider-specific behavior behind adapter boundaries and previews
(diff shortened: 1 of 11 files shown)


priyam-jain-2002:
Implemented the decided MCP credential boundary.

- args remains available for ordinary runtime arguments.
- Manifest load and package validation now reject credential-bearing argument options (for example --token, --api-key=..., Authorization:...) and values matching Lineage’s canonical high-confidence secret patterns.
- Authentication remains declarative through auth: receiver; actual credentials stay in receiver-local environment/provider configuration.
- Added regression coverage for normal args, credential-bearing flags/options, a GitHub-token-shaped literal, and manifest-load rejection.

Pushed as commits 158e35c, 3e07a94, c894569, and 6f3d96d. PR Policy is green; the Test workflow is still running.

Rushilkr77 (approved):
Reviewed against #245's acceptance criteria and built/tested locally at head `6f3d96d` (`go build ./...`, `go vet ./...`, `go test ./internal/packages/... ./internal/cli/... ./internal/runtime/...` all pass).

- Manifest surface (`dependencies.mcp`) validates transport-specific shape correctly: stdio requires `command`/rejects `url`, remote transports require an absolute https URL with no embedded credentials and reject `command`/`args`, unknown fields are rejected up front via `allowedMCPFields` before the typed decoder can silently drop them.
- Credential boundary is properly enforced two ways — `credentialBearingMCPArg` catches flag-shaped credential options, and `validateMCPArgs` also runs args through the same `matchesSecretContent` used by the canonical secret scanner (good: this reuses the canonical pattern set rather than hand-copying it, avoiding the divergence problem flagged on #228).
- `network` capability declaration is checked against the MCP host — a remote server can't be declared without also declaring the capability.
- Provider adapters fail closed correctly: `ValidateMCPProviderSupport` refuses to run when a package has MCP deps and the provider can't materialize them yet, rather than silently dropping the dependency (satisfies that acceptance criterion directly).
- `inspect`/`validate` (both text and structured/YAML output) surface name, transport, URL, and whether auth is receiver-provided — no path for a credential to appear since none can be represented.
- Tests cover the shapes I'd want to see: valid stdio/remote, credential-bearing flag/option, secret-shaped literal (`ghp_...`), duplicate server names, provider-refusal.

Correctly deferred, per the PR's own notes: provider-native MCP config merging (needs its own adapter-boundary design) and threading this into #200's resolved dependency-state model — both flagged as real follow-ups rather than silently out of scope.

One non-blocking note: the branch is reported as behind `develop` — please rebase before merge, just to keep CI honest against current head.

Approving.

priyam-jain-2002 (commented):
Maintainer update: synced current `develop` into this branch at `21e34c1` to clear the stale-base/merge-conflict block. The feature changes themselves were preserved; the sync brings in #103's behavioral-model changes from `develop`. Fresh Test and PR Policy workflows are running on the updated head. Please re-confirm approval on the current head once CI is green.

priyam-jain-2002:
Author update: the branch is now synced with current `develop`, and the fresh Test and PR Policy workflows are green on `21e34c1`. The existing approval is also attached to this current head. From my side this is ready for the normal review/merge flow.

Rushilkr77 (approved):
Checked out `21e34c1` locally and ran `go build ./...`, `go vet ./...`, and `go test ./internal/packages/... ./internal/cli/... ./internal/runtime/...` — all pass.

Logic holds up: credential-bearing arg regex, secret-content scan, https-only/no-embedded-creds/host-in-capabilities.network checks for remote transports, and the fail-closed unknown-field rejection in `validateMCPManifestFields` are all correct and covered by tests.

Two non-blocking notes for a follow-up:
- `gofmt -l` flags `internal/cli/structured.go`, `internal/packages/mcp.go`, `internal/packages/manifest.go` (misaligned struct tags / stray blank line). No gofmt gate in CI so this won't fail, but worth a quick `gofmt -w` pass.
- `ValidateMCPProviderSupport` runs inside `BuildPlan` before the `--dry-run` branch in `runProvider`/`runWorkflow`, so `lineage run <provider> --dry-run` on a package with MCP deps errors out instead of previewing the plan (`lineage inspect` still shows the deps fine). Intentional fail-closed behavior per the README, but dry-run failing feels like a rough edge worth revisiting later.

Approving — nothing here blocks merge.

PR #272, opened 2026-09-09
docs(packages): define content-addressed distribution contract

priyam-jain-2002:
## Summary

Define the authoritative content-addressed package-storage and install-completeness contract for the lightweight distribution goal.

ADR 0017 builds on the existing snapshot store without conflating enable-time snapshots with installed-package state. It specifies immutable asset identity, a versioned content manifest, verified object-store semantics, atomic complete-install references, migration from current archive installs, and boundaries for #266–#268.

## Linked Issue

Closes #265

- [x] The linked issue is assigned to me.
- [x] This PR targets `develop`, or it is a release-tracked promotion into `main`.
- [ ] This PR is not ready for review until the linked issue and test plan are complete.

## Package Impact

- [x] Package format or manifest behavior
- [ ] Package discovery or enablement
- [ ] Provider launch/shim behavior
- [ ] Setup or permission flow
- [x] Documentation only
- [ ] Tests only

## Merge Method

- [x] Squash merge for an ordinary PR into `develop`.
- [ ] Merge commit for a release promotion into `main`.
- [ ] Merge commit for a `main` to `develop` sync-back.

## Safety

- [x] This change does not export secrets, credentials, private prompts, or machine-local state.
- [x] Package inputs are treated as untrusted.
- [x] Path handling prevents traversal outside the intended workspace/package root where applicable.
- [x] Provider-specific logic stays behind an explicit boundary.

## Verification

Relevant test cases:

- ADR examples were checked against the existing package manifest, deterministic digest, archive pull/import, snapshot, and materialization flows.
- This is a documentation-only contract PR; implementation tests belong to #266, #267, and #268.

```text
git diff --check
go test ./...
```

`git diff --check` passed locally. `go test ./...` could not run because the execution environment has no `go` binary; repository CI must run it.

If this PR does not need automated tests, explain why:

- No runtime behavior changes in this PR. The ADR defines the implementation and test obligations for its child issues.

## Notes For Reviewers

Please focus review on the distinction between a verified cached object and a complete offline-usable installed release, and on the legacy archive migration boundary.

PR #275, opened 2026-09-12
Add versioned behavioral model for compiled workflows (#103)

### LINKED ISSUE #103: Define the behavioral workflow model for compiled packages

Rushilkr77:
## Summary

Adds internal/model.BehavioralModel, a versioned, provider-neutral representation of a compiled workflow built from internal/inventory's evidence. It sits between the evidence inventory (#203) and agent-assisted analysis (#104) in the pipeline ADR 0016 describes.

## Linked Issue

Closes #103

## Package Impact

## Merge Method

## Safety

## Verification

Relevant test cases:

- TestValidateAcceptsValidModel, TestValidateRejectsBadSchema, TestValidateRejectsEmptySteps
- TestValidateRejectsStalePerCitationEvidence, TestValidateRejectsStaleSourceInventoryDigest
- TestValidateRejectsEvidenceNotInInventory, TestValidateRejectsClaimWithNoEvidence, TestValidateRejectsDuplicateClaimValue
- TestValidateRejectsEmptyStepID, TestValidateRejectsEmptyClaimValue, TestValidateRejectsEmptySetupNeedPath, TestValidateRejectsEmptyGateID, TestValidateRejectsEmptyDecisionID
- TestValidateRejectsRefNamingWrongField, TestValidateAcceptsModelLevelDecision, TestValidateAcceptsFieldLevelRefWithNoClaimYet
- TestLoadInventoryRoundTrip, TestLoadInventoryDefaultsAbsentSchema, TestLoadInventoryRejectsUnknownSchema
- TestComputeInventoryDigestStableAcrossOrder, TestComputeInventoryDigestChangesOnContentChange, TestNewSkeletonFromInventoryProducesModelLevelDecisions
- TestMessyWorkflowRepresentableWithoutProviderFields

```text
go test ./...
```

## Release Tracking

Not applicable for this `develop` PR.

## Notes For Reviewers

Design decisions (recorded in ADR 0018):

1. Evidence linking: every Claim carries its own EvidenceRef, not one shared per step.
2. No synthetic IDs where a field already has a natural key: Claim.Value and SetupNeed.Path double as identity. Gate keeps a real ID since its only content is free prose.
3. Decision.Refs addresses model-level, step-level, field-level, or claim-level ambiguity via one Ref type.
4. SourceInventoryDigest pins the whole inventory snapshot; Validate checks it alongside per-claim evidence.

Since the initial review: evidence drift (whole-inventory and per-citation) is now a Validate error, not a note, with regression coverage. Identity fields (Step.ID, Claim.Value, SetupNeed.Path, Gate.ID, Decision.ID) are now required non-empty, with regression coverage. ADR renumbered 0017 to 0018 to clear a collision with #272.


docs/decisions/0017-behavioral-model-is-a-versioned-evidence-linked-schema.md
```diff
@@ -0,0 +1,73 @@
+# 0017 Behavioral Model Is A Versioned, Evidence-Linked Schema
```
priyam-jain-2002: ADR number `0017` is already allocated by open #272 for the package content-addressing contract. These records cannot both be `0017`; please renumber this behavioral-model ADR to the next available number and add it to `docs/decisions/README.md` in the same change.

priyam-jain-2002 (changes requested):
Blocking review — two contract issues need correction before this schema becomes a foundation for #104/#106:

1. `Validate` records both a changed `SourceInventoryDigest` and stale cited evidence as notes, yet `Passed()` remains true. A consumer that gates compilation on `Passed()` can therefore compile a model no longer supported by the supplied evidence. Evidence drift must be a validation error, or the API must expose a separate validity state that every compiler is required to reject; the current comments and follow-up do not establish that invariant.

2. The ADR-number collision is an immediate documentation conflict with #272. Renumber and index it.

Please add regression tests for stale whole-inventory and per-citation evidence being non-compilable, then request another review. The overall provider-neutral boundary and explicit decisions are directionally right; this is about preserving the no-invention/evidence contract under source drift.

Rushilkr77:
Both fixed on the latest push:

1. Evidence drift (SourceInventoryDigest mismatch and per-citation digest mismatch) is now a Validate error, not a note. Passed() is false whenever evidence has drifted at either level. Added TestValidateRejectsStalePerCitationEvidence and TestValidateRejectsStaleSourceInventoryDigest as regression coverage, and documented the invariant in ADR 0018 and in Validate/Passed's own comments.

2. Renumbered the ADR to 0018 to clear the collision with #272's 0017, and added the index entry to docs/decisions/README.md.

internal/model/model.go
```diff
@@ -0,0 +1,163 @@
...
+	Steps []Step `json:"steps"` // order is semantic — the workflow's actual sequence, not sorted
+
+	// Decisions holds every unresolved or ambiguous piece of behavior,
+	// both model-level (Refs empty) and step/field/claim-scoped (Refs
+	// naming precise anchors). Sorted by ID for deterministic output.
+	Decisions []Decision `json:"decisions"`
+}
+
+// Step is one ordered unit of the workflow.
+type Step struct {
```
priyam-jain-2002, line 52: `Validate` detects duplicate step IDs but accepts one empty ID. That is not a valid stable identifier: a `Ref{StepID: ""}` means model-level by design, so no Decision can address that step, and #106 cannot map it deterministically into a workflow step. Require non-empty IDs (and likewise validate the other declared identity fields—at least Decision.ID and Claim.Value) with regression tests.

priyam-jain-2002 (changes requested):
The prior blockers are fixed: ADR 0018 is correctly indexed, and evidence drift now fails `Validate` with direct regression coverage.

One blocking schema-validity gap remains: identity fields are documented as stable/natural keys, but a model with an empty Step.ID currently passes validation. That breaks the stated Ref semantics and downstream mapping contract. Please validate required identity fields before this becomes the input to #104/#106, with tests for empty step/decision IDs and claim values.

Separately, current PR Policy is failing even though Test passes. Its log shows the PR body needs a `Relevant test cases:` section and a `Merge Method` section with squash merge selected. Please update the PR description so branch protection can pass. One non-blocking cleanup: `LoadInventory`’s comment says an explicit zero schema defaults, while its code correctly rejects it.

Rushilkr77:
All three addressed:

1. Empty identity fields now rejected. Step.ID, Claim.Value, SetupNeed.Path, Gate.ID, and Decision.ID must all be non-empty — the prior check only caught a second occurrence via the duplicate map, so a single empty ID passed through silently. Added registerIdentity, checking emptiness before the duplicate check, applied uniformly across all five fields, with one regression test each. Documented in ADR 0018's Decision section, not just in code comments.

2. PR description now follows the template (Summary/Linked Issue/Package Impact/Merge Method/Safety/Verification/Release Tracking/Notes For Reviewers), with Relevant test cases and squash merge selected. PR Policy is green.

3. Fixed the LoadInventory comment: it said an absent or explicit-zero schema both default to CurrentSchema, but the code has always correctly rejected an explicit zero. Comment now matches behavior.

Also added a CHANGELOG entry under Unreleased for this change.

Ready for another look.

priyam-jain-2002 (approved):
Re-reviewed the latest head. The prior blockers are resolved: required identity fields are validated with direct regression coverage; ADR 0018 and its index are consistent; the PR contract now satisfies policy; and both Test and PR Policy pass.

Approved. Non-blocking follow-up: the model comment says Decisions are sorted by ID for deterministic output, but neither validation nor NewSkeletonFromInventory enforces that ordering. Define canonical ordering when a serialized model format/loader is added, rather than relying on callers.

PR #276, opened 2026-09-12
fix: reject duplicate skill exports across enabled packages

### LINKED ISSUE #246: Detect duplicate skill exports across enabled packages

MsfPablo:
Fixes #246.

## What

`ValidateDependencies` now builds a map from exported skill name → providing packages and fails when two or more enabled packages export the same skill name, instead of letting later resolution order silently pick a winner.

## Error behavior

- Exactly two providers: `cannot enable package set: skill "research" is exported by both "bar" and "foo"`
- Three or more: `... is exported by multiple packages: "alpha", "mid", "zeta"`
- Skills and providers are sorted, so the first conflict reported — and the message itself — never depends on package/discovery iteration order.

## Compatibility

A valid enabled set with unique skill names validates exactly as before, and a package requiring a skill it exports alone still passes (no self-conflict).

## Verification

- `go test ./... -race -count=1`: all pass, including four new cases — duplicate export fails with both names, order-independence of the message, triple export lists all providers sorted, and the self-requirement non-conflict.

Assisted-by: Claude <noreply@anthropic.com>

### DIFF
--- a/internal/packages/dependencies.go
+++ b/internal/packages/dependencies.go
@@ -1,26 +1,70 @@
 package packages
 
-import "fmt"
+import (
+	"fmt"
+	"slices"
+	"sort"
+	"strings"
+)
 
 // ValidateDependencies checks that every package's Requires.Skills is
-// satisfied somewhere across pkgs — its own skills, or another package's.
+// satisfied somewhere across pkgs — its own skills, or another package's —
+// and that no skill name is exported by more than one enabled package.
 // It's meant to run against the full set of packages that will be enabled
 // together (not one package in isolation), since a required skill can
 // legitimately be provided by a different enabled package.
+//
+// A skill exported by two enabled packages is ambiguous: whichever
+// definition wins later resolution order would silently satisfy the
+// dependency, so validation fails instead of choosing (#246).
 func ValidateDependencies(pkgs []Package) error {
-	available := map[string]bool{}
+	providers := map[string][]string{}
 	for _, pkg := range pkgs {
 		for _, skill := range pkg.Skills {
-			available[skill] = true
+			if !slices.Contains(providers[skill], pkg.Manifest.Name) {
+				providers[skill] = append(providers[skill], pkg.Manifest.Name)
+			}
 		}
+	}
+
+	for _, skill := range sortedKeys(providers) {
+		pkgs := providers[skill]
+		if len(pkgs) < 2 {
+			continue
+		}
+		sort.Strings(pkgs)
+		if len(pkgs) == 2 {
+			return fmt.Errorf("cannot enable package set: skill %q is exported by both %q and %q", skill, pkgs[0], pkgs[1])
+		}
+		return fmt.Errorf("cannot enable package set: skill %q is exported by multiple packages: %s", skill, quotedJoin(pkgs))
 	}
 
 	for _, pkg := range pkgs {
 		for _, required := range pkg.Manifest.Requires.Skills {
-			if !available[required] {
+			if _, provided := providers[required]; !provided {
 				return fmt.Errorf("package %q requires skill %q, which is not provided by any enabled package", pkg.Manifest.Name, required)
 			}
 		}
 	}
 	return nil
+}
+
+// sortedKeys returns providers' skill names in sorted order so the first
+// conflict reported — and therefore the error message — is independent of
+// package/discovery iteration order.
+func sortedKeys(providers map[string][]string) []string {
+	skills := make([]string, 0, len(providers))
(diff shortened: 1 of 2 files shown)


internal/packages/dependencies.go
```diff
@@ -1,26 +1,70 @@
...
 // together (not one package in isolation), since a required skill can
 // legitimately be provided by a different enabled package.
+//
+// A skill exported by two enabled packages is ambiguous: whichever
+// definition wins later resolution order would silently satisfy the
+// dependency, so validation fails instead of choosing (#246).
 func ValidateDependencies(pkgs []Package) error {
-	available := map[string]bool{}
+	providers := map[string][]string{}
 	for _, pkg := range pkgs {
```
priyam-jain-2002, line 22: This deduplicates providers by `pkg.Manifest.Name`, so two distinct enabled packages with the same name (different version, source, or path) are collapsed and the ambiguous export passes validation. That is still an order-dependent provider collision downstream. Track distinct resolved package identity for conflict detection, and render an actionable display identity (`name@version`, with path only when still needed) rather than treating the manifest name as identity.

priyam-jain-2002 (changes requested):
Blocking: this duplicates the existing contribution in #254, which already has an unresolved correctness finding. Please do not maintain two implementations for #246. Choose one PR to repair, close the other, and carry forward the open review requirement: only the same resolved package/path may be deduplicated unless the deduplicated set is also what materialization consumes. The selected PR must detect different copies/versions that share a manifest name and test that case.

Rushilkr77 (changes requested):
Confirming the blocking finding above still stands — no commits since 2026-09-12, and the duplicate hasn't been resolved. Checked the current diff directly: this PR's own error-message example (`both "bar" and "foo"`) is exactly the ambiguous-identity case that was found and fixed in #254 (name@version widening when manifest names collide), and this branch has neither that fix nor #254's duplicate-package dedup. It's also currently failing the `PR Policy` check.

Please close this PR in favor of #254, which already has the identity-widening fix and more review history against #246 (there's one unresolved dedupe/materialization finding there I'm tracking separately). If there's something in this branch's approach that #254 is missing, raise it as a comment on #254 instead of maintaining two competing implementations.

Issue #266, opened 2026-09-02
Package distribution: implement local content-addressed storage and cross-package deduplication


priyam-jain-2002:
## Problem

Once the content-addressed contract in #265 is settled, Lineage needs a concrete local object store that makes digest identity real. Today package bodies can remain package-shaped copies; that prevents reliable cross-package deduplication and gives later install/update code no canonical place to resolve verified immutable content.

## Goal

Implement the local content-addressed storage layer defined by #265 so identical package assets are stored once per digest and can be safely resolved by any installed package/version.

## Maintainer triage

- **Priority:** P1 / high
- **Readiness:** blocked by #265
- **Dependency:** #265

## Required behavior

- Store immutable package content under the digest identity defined by #265.
- Reuse an existing verified object when another package/version references the same digest.
- Use safe/atomic writes so interrupted writes cannot create an object that appears valid.
- Verify content before accepting it into the store and before materialization when required by the contract.
- Distinguish cache miss, corruption, and metadata/manifest mismatch with actionable errors.
- Keep package installation/reference metadata separate from immutable content bytes.
- Preserve compatibility with existing installed package state during migration.

## Safety and filesystem constraints

- Content lookup must not permit path traversal from manifest-controlled logical paths or digest strings.
- Symlink/file-mode behavior must follow #265; do not infer additional semantics in this implementation.
- A failed write must not clobber an already verified object.
- Concurrent attempts to store the same digest must converge safely on one valid object.
- Never trust filename/path identity in place of recomputing/verifying the digest where the contract requires verification.

## Tests

At minimum cover:

1. two different packages sharing byte-identical content -> one stored object;
2. two versions sharing N-1 objects -> shared objects remain single-copy;
3. repeated installation is idempotent;
4. corrupt stored object is detected;
5. interrupted/partial write is never observed as valid;
6. concurrent same-digest writes are safe;
7. malicious logical path/digest input cannot escape the store;
8. existing package/install state remains readable through the migration path.

## Non-goals

- Do not implement registry delta-fetch policy here beyond the local APIs needed by #267.
- Do not implement advanced cache eviction/GC unless required by #265 for correctness.
- Do not put provider-specific materialization logic into the CAS.
- Do not add token estimation in this issue.

## Acceptance criteria

- [ ] The CAS implementation follows #265 without introducing new digest semantics.
- [ ] Byte-identical content referenced by multiple packages/versions exists once in the local object store.
- [ ] Reads return only verified content or an explicit failure.
[...]

Issue #268, opened 2026-09-02
Package inspection: report package bytes and estimated context weight


priyam-jain-2002:
## Problem

Lineage can distribute and materialize package context, but authors and operators cannot currently see the practical weight of a package before enabling it. The existing product principle is progressive disclosure: lightweight metadata/stubs should be cheap to expose, while full bodies are loaded only when needed. That principle is difficult to validate if Lineage cannot report the approximate cost of package assets.

This issue adds observability for #264 after the storage/manifest semantics are fixed by #265. It must not repurpose unrelated portability fields simply because they are currently empty.

## Goal

Expose package storage/transfer weight and estimated context/token weight through package inspection/diagnostics using a dedicated, semantically correct schema.

## Maintainer triage

- **Priority:** P1 / high
- **Readiness:** blocked on #265 for canonical manifest/asset metadata; implementation can proceed independently of registry delta transfer once that contract is stable
- **Dependency:** #265

## Proposed model

Represent weight as explicit metadata/derived diagnostics rather than provider-independent truth. At minimum distinguish:

- stored/content bytes;
- stub/manifest-facing bytes where applicable;
- full-body bytes;
- estimated stub-context tokens;
- estimated full-body-context tokens;
- estimator/tokenizer identity and version;
- whether a value is exact (bytes) or estimated (tokens/context).

The concrete type names should follow repository conventions, but `PortabilityReport.ParameterizedValues` should not be overloaded unless #265 establishes that this data is genuinely part of that field's semantics.

## CLI behavior

`lineage inspect <pkg>` should expose enough information to answer:

- How large is this package's unique required content?
- How much of it is already present locally versus missing, when local state is available?
- What is the approximate context cost of each skill/workflow/policy body?
- What is the lightweight/stub cost versus the full-body cost where progressive disclosure exists?
- Which tokenizer/estimator produced the token estimate?

`lineage doctor` may warn on missing/invalid estimator metadata or unexpectedly large context surfaces, but should not invent universal hard limits in this issue.

## Estimation constraints

- Never present estimated token counts as exact/provider-neutral facts.
- Estimates must be reproducible for a named estimator/tokenizer version.
- If provider-specific estimators are later supported, preserve a generic fallback and make the selected estimator visible.
- Do not require a network/model API call merely to inspect package weight.

## Tests

Cover at minimum:

1. exact byte accounting from manifest/content metadata;
2. deterministic token estimate for a fixed estimator fixture;
[...]

Issue #267, opened 2026-09-02
Package distribution: fetch and update packages by missing content digests


priyam-jain-2002:
## Problem

After #265 defines the content-addressing contract and #266 provides the local CAS, install/update/sync still needs to stop treating a package version as an indivisible blob. A package update that changes one asset should not re-transfer unchanged bodies that are already present and verified locally.

This issue turns content-addressing into actual distribution behavior.

## Goal

Make package resolution and transfer digest-aware: resolve the target manifest, determine which referenced immutable objects are already present and valid in the local CAS, and transfer only the missing objects required to make the installed package complete.

## Maintainer triage

- **Priority:** P1 / high
- **Readiness:** blocked by #265 and #266
- **Dependencies:** #265, #266

## Required behavior

For install/update/sync:

1. resolve and validate the target package manifest;
2. derive the required digest set;
3. ask the local CAS which objects are present and valid;
4. fetch only missing objects;
5. verify every fetched object against the manifest contract;
6. atomically record the target package/version as installed only when its required object set is complete;
7. allow provider materialization to resolve the completed package entirely from local verified content.

A version transition such as:

```text
v1 -> {A, B, C}
v2 -> {A, B, D}
local CAS -> {A, B, C}
```

must require transfer of `D`, not `A`, `B`, and `D`.

## Install versus lazy resolution

Normal `lineage add`/install semantics must remain complete and offline-usable. If a deferred/lazy mode is introduced, it must be explicit in CLI/state and must never masquerade as a complete installation.

The default implementation should prefer **missing-digest synchronization**, not a new binary-delta protocol.

## Registry/API considerations

- Reuse the existing registry/package API where practical; do not freeze a broader self-hosting/CDN contract here.
- If the current registry endpoint cannot fetch immutable content independently, add the narrowest content-fetch endpoint/shape required by #265.
- Private/org authorization from #239 remains orthogonal: authorization determines whether content may be fetched; digest semantics determine what content is fetched.
- Avoid leaking unauthorized private-package existence through digest-fetch behavior.

## Failure behavior

- A failed or interrupted transfer must not mark the target version complete.
- Digest mismatch fails the install/update and does not poison the CAS.
- Network failure may leave verified newly fetched objects cached, but package state must accurately report incompleteness.
- Retrying should reuse any verified objects already present.
- Existing installed versions must remain usable if an update attempt fails.

## Tests

At minimum cover:

1. fresh install fetches the required unique digest set;
[...]

PR #279, opened 2026-09-18
feat(snapshot): add verified package content store

priyam-jain-2002:
## Summary

Implements #266's local content-addressed package store using the ADR 0017 contract. Immutable source files are stored once per SHA-256 digest, complete installed-release references are committed only after every object and the package digest verify, and releases can be materialized through the common store boundary.

The object writer now publishes same-directory temporary files with atomic no-clobber semantics, re-verifies concurrent winners, and rejects corrupt cache entries. Legacy installed directories remain usable and can be migrated explicitly without being changed in place.

## Linked Issue

Closes #266

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- Concurrent writers converge on one verified object without overwriting a winner.
- Corrupt and missing objects are distinct, and corrupt cache entries are rejected.
- Two packages that share identical content reuse the same object ID.
- Incomplete or digest-mismatched releases do not become installed or replace a prior complete release.
- Legacy installs migrate to a complete reference and materialize through the verified store.
- A symlinked package manifest is rejected before admission to the store.

```text
go test ./...
go vet ./...
go build ./...
```

## Notes For Reviewers

This is intentionally local storage and resolution only. Registry missing-digest transfer belongs to #267, while package/context-weight reporting belongs to #268. The main review focus is the atomic no-clobber writer and the distinction between verified cached objects and a complete installed release.


Rushilkr77 (approved):
Reviewed the content-addressed store implementation for #266. Traced the concurrent-write path (writeBlobNoClobber + re-verify via getBlob on link-exists), CommitRelease's verify-before-commit gate, and SafeJoin/symlink handling — all correctly enforce the atomicity and integrity requirements in #266's acceptance criteria. Interface (ObjectAvailability, WriteObject, CommitRelease) lines up well with #267's digest-aware fetch needs.

One non-blocking note: verifyContentManifestObjects re-implements packages.ComputeDigest's hash ordering independently rather than reusing it — currently equivalent but worth extracting to a shared helper to avoid future drift.

CI green.

PR #283, opened 2026-09-18
feat(distribution): sync packages by missing digest

priyam-jain-2002:
## Summary

Implements digest-aware package distribution for #267. Content-manifest-aware registries now send only objects missing from the verified local CAS, then atomically commit the release after every object and the reconstructed package digest validate.

`lineage add`, `lineage package pull`, and archive import use the shared distribution boundary. Existing archive-only registries remain supported: their verified archive installs are migrated into the CAS and recorded as complete releases. ADR 0017 documents the optional `contentManifest` and same-origin `objectPathTemplate` response fields for registry implementations.

## Linked Issue

Closes #267

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- Fresh content-manifest install fetches each unique object once and remains materializable after the registry is offline.
- Two packages with shared content reuse the cached object instead of downloading it twice.
- A version update downloads only changed content while retaining unchanged objects.
- A partial transfer does not commit a release; retrying reuses verified objects from the first attempt.
- Incorrect object bytes are rejected before they enter the CAS.
- Archive-only registry responses still install and register complete local releases.
- Existing add, pull, import, package, snapshot, and full-suite regression coverage passes through the shared distribution path.

```text
go test ./...
go vet ./...
go build ./...
```

## Notes For Reviewers

The registry server is intentionally not part of this repository. This client accepts the optional ADR 0017 metadata capability when a registry provides it and falls back to the deployed archive API otherwise. Please focus review on the failure boundary: object bytes are size-checked and hashed before storage, cached corruption fails closed, and an installed-release reference is committed only after all required objects verify.


Rushilkr77 (approved):
Reviewed the registry-sync client (ADR 0017) on top of #266's already-merged CAS, with extra attention to cross-component interactions given the blast radius of a distribution-layer change:

- Grepped every caller of the old \`packages.Pull\`/\`packages.Import\` — nothing bypasses the new \`distribution\` wrappers, so every entry point (\`add\`, \`package pull\`, \`package import\`) now registers a release in the CAS.
- Confirmed \`packages.ComputeDigest\` (used for the \`importAddSource\` idempotency check) and \`ContentManifest.PackageDigest\` resolve to the exact same function via \`Discover\` — no digest drift between the legacy and content-manifest paths.
- \`objectPathTemplate\`/\`{digest}\` from the registry: digests are regex-locked (\`^sha256:[0-9a-f]{64}$\`) before touching a URL, and the template is same-origin-only (rejects \`//host\` protocol-relative bypass) — no SSRF surface.
- Built a throwaway fake content-manifest registry and ran the real client against it end-to-end (isolated \`LINEAGE_HOME\`): malformed manifest fails closed pre-fetch; a tampered \`package_digest\` still gets every object individually verified and stored, but \`CommitRelease\` correctly refuses to advertise the release as complete (defense in depth beyond per-object hashing); happy path produces correct dedup'd objects, correct release record, correct materialized files; re-pull of the same ref cleanly rejects instead of silently overwriting; the resulting package then \`validate\`/\`enable\`/\`doctor\`/\`run --dry-run\` exactly like any other package.
- \`go build\`, \`go vet\`, full \`go test ./...\` all pass.

One non-blocking gap for a follow-up: \`lineage doctor\`'s "snapshot store" check only covers the older ADR-0014 enable-time snapshots and has zero visibility into \`package-releases/\` (confirmed live — doctor reported 0 manifests right after a successful CAS import that did write a release record). Pre-existing since #266, but this PR is what first makes it reachable from \`lineage add\`/\`pull\`, so doctor currently can't detect a corrupted CAS-installed release. Will file separately.

Approving.

PR #286, opened 2026-09-18
feat(inspect): report package weight

priyam-jain-2002:
## Summary

Adds a dedicated package-weight diagnostic to `lineage inspect` and `lineage inspect --yaml`.

- Derives exact per-asset and aggregate bytes from the canonical content manifest.
- Separates manifest-facing stub weight from full-body weight.
- Reports deduplicated local CAS bytes as verified, missing, or corrupt without changing logical package-size semantics.
- Uses a named offline `bytes-per-token v1` heuristic for text assets; non-text context estimates are explicitly unavailable.
- Documents that context values are estimates, not provider-independent runtime facts.

## Linked Issue

Closes #268

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- exact logical bytes and deterministic estimator metadata;
- per-asset text and non-text estimates;
- explicit unavailable estimate for non-text content;
- stub versus full-body totals;
- shared CAS-object deduplication for local storage;
- human and YAML inspection output.

```text
go test ./...
go vet ./...
go build ./...
git diff --check
```

- Automated tests were added.

## Release Tracking

Complete this section only for PRs targeting `main`.

- [ ] Release-worthy main promotion
- [ ] Non-release housekeeping

Planned version:

`v0.0.0`

Release classification:

`patch | minor | major`

Release notes:

What changed:

- Not applicable: this PR targets `develop`.

What is experimental:

- Not applicable.

What is outside scope:

- Not applicable.

Safety and compatibility notes:

- Adds an optional `weight` field to structured inspection output; existing fields are unchanged.

Post-merge tag plan:

- [ ] Create and push an annotated SemVer tag after merge to `main`.
- [ ] Publish the matching GitHub Release and artifacts.
- [ ] Sync the resulting `main` merge commit back into `develop`.

Non-release housekeeping reason:

- Not applicable.

## Notes For Reviewers

The generic estimator is deliberately offline and deterministic. It makes no claim about a provider's actual tokenization or runtime prompt consumption. A future provider-specific tokenizer can add another named estimator without altering the exact byte or CAS accounting.


Rushilkr77 (changes requested):
Checked out the branch and ran it: `go build`, `go vet`, full `go test ./...` pass, and the new files are gofmt-clean (this also fixes the pre-existing gofmt drift in `structured.go`). I also ran `lineage inspect` live in an isolated `LINEAGE_HOME` against edge-case packages and compared against `develop`. Corrupt-object reporting, deduplicated-vs-logical byte accounting, and the `--yaml` shape all behave as documented. Three things need fixing first.

**1. Blocking: `inspect` now hard-fails on packages it displays fine on `develop`.**
`runInspect` (`internal/cli/cli.go`) now calls `snapshot.BuildContentManifest(pkg.Path)` and returns on any error. `readRegularPackageFile` rejects anything that isn't a regular file, so a package whose `lineage.yaml` is a symlink inspects normally on `develop` (and `lineage package validate` even reports PASS for it) but on this branch prints only `read lineage.yaml for content manifest: not a regular file` and nothing else. `inspect` is the receiver's read-only look before `enable`; the weight block is supplementary and should never take the rest of the report down with it. Please degrade instead: print the normal report and either omit `weight` (nil in YAML) or emit a one-line warning when the manifest can't be built. Add a regression test for the symlinked-manifest case. (Separate pre-existing gap: `validate` passing a manifest that can't produce a content manifest looks worth its own issue.)

**2. The headline estimate is unavailable for ordinary packages.**
`isTextAsset` keys only off `MediaType`, and `contentMediaType` only recognises `.md`, `.mdc`, `.yaml`, `.yml`, `.json`, case-sensitively. A `notes.txt`, `.sh`, `.py`, `.toml` or an uppercase `README.MD` is classed non-text, and because `addWeight` poisons the bucket, one such file makes `full_body_context_tokens` `unavailable` for the entire package. Reproduced with `skills/s/SKILL.md` + `references/notes.txt` + `references/README.MD`. Suggest treating an asset as text when the extension is a known text one (case-insensitive) or, better, when the bytes are valid UTF-8 with no NUL, and add a test with a `.txt` asset.

**3. Empty bucket reported as unavailable with an empty reason.**
A manifest-only package prints `full_body_context_tokens: unavailable ()`. An empty bucket has no non-text assets, so it should be `~0` (Available true, Tokens 0). `addWeight` never marks a bucket available until it sees a text asset, so the zero value leaks through as "unavailable" with no reason.

**Smaller notes (non-blocking):**
- `ObjectAvailability` maps every non-`IsNotExist` read error to `ObjectCorrupt`, so a permissions problem on the objects directory would be reported as corrupt bytes. Worth distinguishing, or at least documenting.
- `inspect` now reads each package file fully into memory again (`BuildContentManifest`) and re-reads and re-hashes every present CAS object via `ReadObject`, on top of `Discover`'s digest pass. Fine for typical packages, but worth a note for large ones.

Happy to re-review once these are addressed.

priyam-jain-2002:
Addressed the requested changes in 7693aef:

1. Weight calculation is now best-effort. `inspect` retains its normal report when content-manifest construction fails; human output shows a single `weight: unavailable (...)` line and YAML omits `weight`. Added a symlinked-manifest regression test for both modes.
2. Text classification now recognizes common source/config/text extensions case-insensitively, including `.txt` and uppercase `.MD`. Added coverage proving those assets retain a full-body estimate.
3. Manifest-only packages now report an available full-body context estimate of zero.

Verified again with `go test ./...`, `go vet ./...`, `go build ./...`, and `git diff --check`.

Rushilkr77 (approved):
Re-reviewed at `7693aef`. All three items from my earlier review are fixed and verified live in an isolated `LINEAGE_HOME` (not just via the new tests). `go build`, `go vet`, full `go test ./...` pass, and the files this PR touches are gofmt-clean.

- **`inspect` regression:** a package with a symlinked `lineage.yaml` now prints its normal report plus `weight: unavailable (read lineage.yaml for content manifest: not a regular file)`; `--yaml` omits `weight` entirely. No hard failure.
- **Text classification:** `.txt` and uppercase `.MD` now keep the full-body estimate (`~11` on my repro).
- **Empty bucket:** manifest-only package shows `full_body_context_tokens: ~0`.

Cross-checked the side effect of the fix: `contentMediaType` also feeds the persisted ADR 0017 content manifest, so I confirmed nothing compares manifests for equality (`CommitRelease` overwrites; `validateContentManifest` doesn't check `media_type`), so previously committed releases aren't affected.

Non-blocking follow-ups (not caused by this PR, feel free to defer):
- The extension allowlist is still brittle: extensionless files (`LICENSE`, `Makefile`) and `.csv` still flip the whole full-body estimate to unavailable.
- A binary blob named `*.txt` gets a made-up token estimate (random bytes named `blob.txt` reported `~111`). A UTF-8/NUL sniff would handle both this and the allowlist gaps.

Approving.

Issue #107, opened 2026-08-22
Emit Claude Code artifacts from compiled package output


priyam-jain-2002:
## Problem statement

Claude Code needs provider-visible context and skill layout that matches its conventions. The compiler should emit Claude-compatible artifacts from the generated package model without copying arbitrary source `CLAUDE.md` content wholesale.

## Scope

- Generate Claude-facing instruction content from the behavioral model and package artifacts.
- Ensure generated skills and workflow summaries materialize correctly through the Claude adapter from #18.
- Preserve ordered workflow intent and operational constraints in Claude-visible text.
- Avoid duplicating or conflicting with the generated section maintained by materialization.

## Acceptance criteria

- A generated package materializes for Claude using the existing #18 path.
- `CLAUDE.md`/Claude-visible context includes active package, workflow, ordered steps, constraints, and references needed to run the workflow.
- Tests verify that generated Claude artifacts are deterministic and do not duplicate materializer-managed blocks.
- The package still passes `lineage package validate` from #31.

## Dependencies

- Depends on provider-neutral artifact generation.
- Builds directly on the Claude materializer adapter from #18.

## Non-goals

- No changes to Claude Code itself.
- No new provider registry behavior.
- No unconstrained rewrite of source `CLAUDE.md`.

## Sequencing

Land after provider-neutral generation and before end-to-end behavioral validation for Claude.

PR #287, opened 2026-09-18
Compile behavioral models into provider-neutral packages (#106)

Rushilkr77:
## Summary

Closes the compile loop for the milestone-11 chain (#102 → #103 → #104): adds `lineage compile <model.json> <source-path> --out <dir>`, which turns a saved, author-reviewed `BehavioralModel` into a provider-neutral Lineage package. It makes no LLM call.

- `lineage analyze --model-out model.json` saves the full model as indented JSON (`--yaml` only ever emitted a lossy summary). `model.ParseModel` reads it back.
- `internal/compile` generates `lineage.yaml`, one `skills/<step-id>/SKILL.md` per step, `workflows/<name>/WORKFLOW.md` (new `packages.SaveWorkflow`, the missing writer counterpart of `LoadWorkflow`), per-step `scripts/` and `references/`, and review-only `references/evidence/<step-id>.md`.
- **Human in the loop:** compile refuses while any `Decision` is unresolved (the rule ADR 0018 and the guide state but nothing enforced). Per blocking decision it prints what it applies to, the `path:line` evidence with digests to copy into a new claim, the JSON path to edit, and the re-run command.
- **Fail closed:** refuses if `model.Validate` fails, a cited file's digest changed, ids/paths are unsafe, or two files would collide (including case-only collisions). The package is built in a staging directory, checked with `packages.Validate` (including the secret scan), then moved into place. `--force` replaces only an existing Lineage package and keeps the old one if the move fails.
- **Tools are classified, not assumed to be files:** an `executable_helper` inventory entry is copied to `skills/<step-id>/scripts/`; a prose-only tool (`gh`, `kubectl`) is not copied and appears in the step's "Requires (install yourself)" section and the report. Lineage installs nothing.
- **Notes (never block):** claim kind vs inventory kind, workspace files no claim or evidence mentions ("not represented in package"), shadowed or unbundled required skills, large/binary files, machine-local paths in copied files, gates described but not enforced, and the Cursor SKILL.md-only limit.
- Generated `SKILL.md` holds the model's summary only, never source text. Every model string is collapsed to one line so a hand-edited or LLM-produced value cannot inject markdown structure into a skill.
- Docs, ADR 0018, ROADMAP, CHANGELOG, README and stale comments updated in this PR.

Plan and an open question about analysis provider access (any API key, subscription CLI, manual): https://github.com/agentic-lineage/lineage/issues/106#issuecomment-5732916893. That is not part of this PR.

## Linked Issue

Closes #106

## Package Impact

- [ ] Documentation only
- [ ] Tests only

## Merge Method

## Safety

## Verification

Relevant test cases:

- `internal/compile`: two-step package passes `packages.Validate` and round-trips `LoadWorkflow`; notes; blocked on decisions with guidance (id, anchor, `path:line`, digest, JSON path); refusal on source drift after inventory, on a stale model, on bad ids, on an empty model, and on case-only step-id collisions; non-executable tool becomes a requirement plus note; determinism (identical `ComputeDigest`); refuses non-empty output and non-package `--force` targets; leaves nothing behind when validation fails (secret in a reference); injected markdown in model strings is neutralized; `--force` replace leaves no staging or backup directories.
- `internal/packages`: `SaveWorkflow`/`LoadWorkflow` round trip and unsafe-name rejection.
- `internal/cli`: argument parsing for `compile` and `analyze --model-out`; end-to-end `analyze --fixture --model-out` → `compile` → `package validate`, plus a blocked compile that creates no output.

```text
go vet ./...
go test ./...
```

## Notes For Reviewers

- Extra scrutiny wanted on: the staging-then-rename write path, `oneLine` sanitization in `internal/compile/render.go`, and where evidence lives (`references/evidence/`, chosen because only `skills, workflows, agents, policies, references, adapters` are digested/exported and the materializer never stages top-level `references/`).
- Per-step `scripts/` and `references/` are correct for Claude/Codex (the materializer copies whole skill dirs) but the Cursor adapter fails closed on any file other than `SKILL.md`; documented, not worked around.
- Known model gaps, surfaced as a note rather than silently dropped: no fields for agents, policies, MCP dependencies or capabilities; entrypoints are #107/#108.
- `gofmt` still flags `internal/packages/manifest.go` and `mcp.go`; both were unformatted before this PR and are left alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


priyam-jain-2002:
Maintainer scope note after reviewing the current diff and the discussion on #106:

**Direction approved. Please keep this PR narrowly on the persisted-model + deterministic compile boundary.** The new provider/access-mode work is now tracked separately in #288.

Concrete decoupling expectations for this PR:

1. **`internal/compile` must remain provider-blind.** No provider names, credentials, model selection, API clients, or agent-CLI execution should enter this package.
2. **`internal/cli/analyze.go` changes in this PR should stop at persistence plumbing.** `--model-out` is in scope because compile needs a complete model. Expanding `resolveAnalysisProvider`, provider selection, auth, model configuration, or data-egress UX is not.
3. **Docs may state the current limitation accurately**, including that live analysis currently uses Anthropic, but should point provider-neutral access work to #288 rather than designing that system inside #287.
4. **Do not make generated package semantics depend on the producer of `model.json`.** A valid BehavioralModel from manual authoring, fixtures, Anthropic, or a future provider must compile identically.
5. **Do not encode missing model concepts by approximation.** Agents, policies, MCP dependencies, capabilities, and provider entrypoints should remain explicit known gaps until their own schema/emitter work lands.
6. **Shared transport is not shared semantics.** Any later OpenAI-compatible transport reuse belongs behind provider adapters in #288, not in this PR.

One concrete item I want addressed before merge: the prior #106 maintainer note requires canonical persisted-model serialization for `Decision` ordering. `writeModelFile` currently marshals the model as provided; please either canonicalize Decisions (for example by ID) before writing or establish another explicit canonical rule, and add a byte-stability test. Package digest determinism does not cover this persisted-model boundary.

The rest of the current compile direction — unresolved-decision refusal, evidence drift checks, staging + `packages.Validate`/secret scan, and explicit notes for unrepresented semantics — is aligned with the issue. Continue on this PR; #288 does not block it unless you discover a change that is genuinely required in the shared BehavioralModel contract.

Rushilkr77:
Thanks, scope noted. Pushed 980ec93 addressing the review:

- **Canonical model serialization (required item):** added `model.MarshalModel` next to `ParseModel`. It writes indented JSON with a trailing newline and Decisions sorted by ID on a copy, so the file never depends on caller iteration order. Steps and claim lists keep their given order (semantic). `analyze --model-out` now calls it. `TestMarshalModelIsCanonicalAndByteStable` checks identical bytes for differently ordered decisions, that the caller's slice is untouched, that step order is kept, and that `ParseModel` → `MarshalModel` round-trips to the same bytes.
- **Provider-blind compile (item 1):** the one note in `internal/compile` that named a specific adapter now says "some provider adapters materialize only SKILL.md…"; the Cursor detail stays in the docs only. `internal/compile` has no provider names, credentials, model selection, clients or CLI execution.
- **Analyze scope (item 2):** `analyze.go` changes remain `--model-out` persistence plumbing only; `resolveAnalysisProvider` is untouched.
- **Docs (item 3):** the guide states the current Anthropic limitation and now points to #288 for provider-neutral access.
- Items 4–6 need no change: compile takes only a validated model, unrepresentable concepts stay explicit "not represented in package" notes, and no transport work is in this PR.

`go vet` and `go test ./...` pass locally.

internal/cli/compile.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	res, err := compile.Compile(m, source, inv)
+	if err != nil {
+		fmt.Fprintln(stderr, err)
+		fmt.Fprintf(stderr, "\nRe-run after editing: lineage compile %s %s --out %s\n", modelPath, source, out)
+		return err
+	}
+	if _, err := compile.Write(res, out, force); err != nil {
```
priyam-jain-2002, line 70: `--force` can currently replace the source workspace itself (or an ancestor containing it). If the source is already a Lineage package, `Compile` reads it successfully and `Write` then renames/removes that same directory, replacing the workspace with only the generated package. Every unrepresented source file is lost.

Please reject overlapping source and output paths before calling `compile.Write` (including ancestor/descendant cases after canonical path resolution), and add a regression test for `source == --out` with `--force`. `--force` may replace an intended output package; it must not turn the analyzed workspace into a destructive output target.

priyam-jain-2002 (changes requested):
One blocking safety issue needs resolution before merge.

internal/cli/compile.go
Rushilkr77: Fixed in 5c8fa64.

`compile.CheckOutputPath(source, out)` now runs first in `lineage compile`, before the model is read or anything is built, and refuses any overlap: the same directory, an ancestor of the workspace, or a directory inside it. Paths are compared after symlink resolution and by file identity (`os.SameFile`), so a symlink alias or differently cased spelling is caught too. An output that does not exist yet is resolved through its nearest existing ancestor. `--force` itself is unchanged; it can only replace an existing Lineage package, and can no longer be aimed at the workspace.

Regression tests:
- `TestCompileForceCannotReplaceTheSourceWorkspace` (CLI): the source is already a Lineage package with an unrepresented file; `--force` with `--out` set to the same directory, its parent, and a directory inside it all fail, and `lineage.yaml` plus `notes/keep-me.md` are still there afterwards.
- `TestCheckOutputPathRejectsOverlap` and `TestCheckOutputPathSeesThroughSymlinks`: equal, dotted, existing and missing inside, parent, grandparent, and symlink alias are rejected; a sibling and an unrelated directory are accepted.

The guide now documents the rule under the author workflow and the refusal list. `go vet` and `go test ./...` pass locally.

priyam-jain-2002 (approved):
Re-review complete. The output/source overlap guard resolves the blocking `--force` data-loss path, including symlink aliases and non-existent descendant targets; the regression coverage and documentation are aligned. CI is green and the PR is cleanly mergeable.

Issue #127, opened 2026-08-22
Safety compliance: define package safety decision taxonomy


priyam-jain-2002:
## Goal

Define the safety compliance taxonomy Lineage uses across package authoring, publishing, inspection, pulling, enabling, and runtime execution.

This is the parent policy decision for deciding what becomes a hard stop, what becomes a warning, and where each safety control must run.

## Why this matters

Lineage packages are not passive archives. They become agent instructions and local files inside a receiver project. A generic secret scan is not enough: the runtime must tell users what is unsafe, what is risky, what is intentionally sensitive, and what is merely informational.

## Pipeline placement

- Publish pipeline: blocks packages that fail hard-stop checks before they enter the registry.
- Inspect pipeline: presents the complete safety posture before enablement.
- Pull/import pipeline: validates package bytes and registry claims before installation.
- Enable/materialization pipeline: shows what Lineage will write into the receiver project.
- Execution/runtime pipeline: governs what active package instructions can influence at run time.

## Examples

Hard-stop candidates:

- `.env`, private keys, GitHub tokens, cloud credentials.
- Archive path traversal or setup paths escaping the project.
- Symlinks or unsupported archive entry types.
- Missing declared workflow skills.
- Registry digest mismatch on pull.
- Unsupported manifest schema.

Warning candidates:

- Personal information included intentionally, such as resume contact info.
- Broad declared filesystem read capability.
- Declared network capability.
- Suspicious but not conclusive instruction patterns.
- No provider compatibility declared.

## Decision needed

Create a reviewed safety policy matrix with columns for check, risk, pipeline stage, default severity, whether it blocks, user-facing copy, and future enforcement path.

## Acceptance criteria

- A committed decision record defines hard-stop vs warning categories.
- Every existing validation check is mapped into the matrix.
- Every planned safety-compliance issue references this taxonomy.
- CLI and registry wording can be derived from the same policy.

## Out of scope

- Implementing every scanner.
- Building a database-backed registry policy engine.

priyam-jain-2002:
**Decision:** Hard-stop is reserved for deterministic/binary checks already enforced elsewhere in lineage — secrets, path traversal/symlinks, missing declared skills, digest mismatch, unsupported schema. PII and instruction-risk detection are **warning-level by default in v1**, never hard-stop — both are heuristic/pattern-based, and escalating them to a hard stop risks false-positive lockouts on legitimate packages.

[...]

Issue #129, opened 2026-08-22
Safety compliance: add PII and personal context detection


priyam-jain-2002:
## Goal

Add PII and personal-context detection for package contents, with reporting across validate, inspect, publish, pull, and enable.

## Why this matters

Secret scanning does not protect personal information. Lineage packages may intentionally include personal data, such as a resume, but the user must know when private or personal context is being packaged or published.

## Pipeline placement

- Publish pipeline: warn or require acknowledgement before publishing packages containing PII.
- Inspect pipeline: show PII categories and affected files.
- Pull/import pipeline: show PII warning before installation or at least before enablement.
- Enable pipeline: show PII warning before materializing instructions.

## Examples to detect

- Email addresses.
- Phone numbers.
- Home or city/location fields.
- LinkedIn, GitHub, portfolio, personal website links.
- Resume employment history and education sections.
- Salary expectations, target companies, recruiter notes, private job-search notes.
- Customer names or confidential project names if represented in package references.

## Strategy

Start with explicit, explainable detectors rather than a black-box classifier. Report category, file path, and reason; never print sensitive matched values unless the user asks for verbose local-only output.

Classify PII as warning by default, not hard stop, because some packages intentionally contain personal context. Escalate to hard stop only for clearly dangerous files or policy violations once the taxonomy decides that.

## Decision needed

Decide PII categories, severity, whether publish requires `--yes`/acknowledgement for PII, and whether registry metadata should expose only summary counts/categories rather than file-level details.

## Acceptance criteria

- PII scan reports categories and file paths.
- Resume package surfaces intentional personal info as a warning.
- Private source notes examples are detected.
- Publish/inspect output clearly distinguishes PII from secrets.
- Tests include both intentional resume PII and accidental private notes.

## Out of scope

- Perfect PII classification.
- Uploading raw matched personal data to the registry.

priyam-jain-2002:
**Decision:** Confirming the proposed default — the listed PII/personal-context categories, warning-only (never hard-stop, per #127), publish requires explicit acknowledgment of any findings. Registry/public surfaces expose **category + file-count only**, never file paths or matched values — extending ADR 0009's "never print the matched value" norm one step further to PII specifically.

Removing `needs:decision`; adding `help wanted`.

Raghavrajvanshi:
## Proposal: PII & Personal-Context Detection (Issue #129 / ADR 0020 rows 2, 13)

[...]

PR #289, opened 2026-09-18
docs: add ADR 0020 safety compliance taxonomy

Raghavrajvanshi:
## Summary

Adds ADR 0020, the parent safety-compliance decision record: a severity
taxonomy (Unsafe/Risky/Sensitive/Informational, collapsing to hard-stop vs.
warning) and a 21-row matrix mapping every existing validation check —
secret scanning, archive/symlink safety, manifest schema, export authority,
workflow/skill references (split into the structural hard-stop case and the
separate non-blocking requires.skills case), digest verification, capability
declarations, provider compatibility, instruction-risk scanning (ADR 0019,
referenced as authoritative, not redefined), and MCP dependency validation —
plus PII detection and capability-breadth classification, which are given a
severity and placement but remain unbuilt.

Docs/policy only. No code changed, no scanners implemented.

## Linked Issue

Closes #127

## Package Impact

- [x] Documentation only

## Merge Method

## Safety

## Verification

Relevant test cases:

- None yet.

```text
go test ./...
```

- Documentation-only change (new ADR + decision-log index entry); no code paths were added or modified. Ran `go build ./...` and the existing `internal/packages`, `internal/materialize`, `internal/runtime` suites locally to confirm nothing regressed, and specifically re-ran `TestValidateRejectsWorkflowWithBrokenStep` and `TestValidateNotesUnsatisfiedRequiredSkillWithoutFailing` to verify the ADR's row 7/7b classification against actual shipped behavior rather than assumption.

## Notes For Reviewers

- Numbering: originally drafted as ADR 0016, renumbered to 0020 after
  0016–0019 landed on `develop` in the meantime.
- Defers to ADR 0019 as-is for instruction-risk scanning's severity split
  rather than re-litigating an already-accepted, already-shipped decision.
- Caught and fixed a real classification bug in review: an earlier draft
  had downgraded the single-package "workflow step references a missing
  skill" check to a non-blocking Risky warning. It's actually a hard-stop
  today (`report.Errors`, verified by `TestValidateRejectsWorkflowWithBrokenStep`).
  Corrected to Unsafe, and split out the genuinely non-blocking
  `requires.skills` case (row 7b) that had been conflated with it.
- Flags that MCP dependency validation (`internal/packages/mcp.go`) is a
  real, already-enforced capability surface with no dedicated ADR yet —
  included here as rows 17–18 with a Follow-Up note recommending its own
  short ADR later, same as 0009/0010/0011 exist for secrets/paths/archives.
- Matrix's Status column strictly means "implemented exactly as classified
  here," not just "related code exists" — e.g. rows 10–12 (capability
  breadth/network severity) are Planned even though the underlying manifest
  fields are already parsed and printed today.


docs/decisions/0020-safety-compliance-taxonomy.md
```diff
@@ -0,0 +1,100 @@
...
+
+This principle does not reopen 0019. Instruction-risk scanning already shipped with its own reviewed severity split (see below) before this ADR was finalized, and that split is adopted here as-is.
+
+### The matrix (21 rows)
+
+Columns: `check`, `category`, `pipeline stage(s)`, `severity`, `blocks?`, `example copy`, `status` (implemented/planned — whether the check exists in code exactly as classified here, not just whether related data exists), `owning reference` (the ADR, implementation issue, code path, or test that backs the row, so documented policy is never confused with what's actually shipped). One row per conceptual check family (e.g. all archive-extraction failure modes are one row), not one row per distinct error string — this is a policy record, not an error-message catalog.
+
+This matrix reflects the state of `develop` as of this ADR's writing, including the MCP-dependency validation (rows 17–18) that merged after the taxonomy's first draft — see the note after row 18.
```
priyam-jain-2002, line 42: Blocking: this currently does not map **every existing validation check** as required by #127. `internal/packages/validate.go` already has at least two distinct hard-stop path checks that are absent from the matrix: provider entrypoint traversal via `validateEntrypoint(...)`, and `setup.files` / `setup.directories` traversal via `SafeJoin(...)`. These are not the same check as archive-entry safety in row 3, and their pipeline placement differs. Please add explicit row(s) for them (with the existing ADR 0010 / implementation references) so this table can actually serve as the complete policy map it claims to be.

docs/decisions/0020-safety-compliance-taxonomy.md
```diff
@@ -0,0 +1,100 @@
...
+| Category | Enforcement class | Blocks? |
+|---|---|---|
+| Unsafe | Hard-stop | Yes |
+| Risky | Warning | No |
+| Sensitive (intentional) | Warning | No |
+| Informational | Info | No |
+
+Sensitive and Informational keep distinct copy/tone from Risky rather than collapsing into one generic "Warning" bucket — "your resume has a phone number in it," "this package wants network access," and "we couldn't verify anything else about it" are different enough in nature that lumping them together would train users to ignore all three.
+
+**Severity is fixed per check, never conditional or escalating** — no scoring engine, consistent with 0009's rejection of an entropy/ML-based engine for secrets. If a specific pattern is bad enough to always block, it gets its own Unsafe-tier check rather than a variable-severity one.
```
priyam-jain-2002, line 32: This invariant is internally inconsistent with the accepted behavior you preserve below in row 14c: `disable_safety` starts as Warn and escalates to Block when co-located with a blocking instruction-risk category. Since ADR 0019 is explicitly not being reopened, the taxonomy cannot simultaneously say severity is *never* conditional/escalating. Please scope this rule to **new checks introduced under this ADR** (or explicitly carve out ADR 0019's accepted conditional rule) so future readers do not interpret the parent policy as contradicting the shipped scanner.

docs/decisions/0020-safety-compliance-taxonomy.md
```diff
@@ -0,0 +1,100 @@
...
+| 13 | Personal info / intentional PII (name, email, phone, address) | Sensitive | Publish, Inspect, Pull/Import | Warning (disclosure) | No | "package includes what looks like personal contact information" | Planned | This ADR |
+| 14a | Instruction risk — exfiltration / silent-destructive action / credential collection | Unsafe | Publish, Inspect, Enable | Hard-stop | Yes | "network action near a credential-shaped noun" / "destructive action without confirmation" | Implemented | `internal/packages/instructions.go` (`ScanForInstructionRisk`); ADR 0019 |
+| 14b | Instruction risk — prompt override / broad local read | Risky | Publish, Inspect, Enable | Warning | No | "content resembles 'ignore previous instructions'" | Implemented | `internal/packages/instructions.go`; ADR 0019 |
+| 14c | Instruction risk — disable-safety language (conditional) | Risky, escalates to Unsafe if co-located with 14a in the same file | Publish, Inspect, Enable | Warning, Hard-stop if paired | Conditional | alone: "disable safety checks" warns; paired with an exfiltration/destructive/credential match in the same file: blocks | Implemented | `internal/packages/instructions.go`; ADR 0019 |
```
priyam-jain-2002, line 62: The instruction-risk mapping is still missing an existing outcome from ADR 0019 / `instructions.go`: `CategoryUnscanned` (oversized/unreadable instruction content). It carries `SeverityWarn`, is intentionally separated from `WarningFindings`, and is surfaced as its own confirmation-relevant blind spot. Because this ADR says every existing check/outcome is mapped, please give `unscanned` an explicit row/category and pipeline placement rather than leaving it implicit under 14a–14c.

priyam-jain-2002 (changes requested):
I cross-checked ADR 0020 against #127, ADRs 0006/0009/0010/0011/0019, and the current validation/runtime code. The direction is coherent, but the matrix is not yet complete enough to be the claimed source of truth: existing hard-stop path validation and the instruction scanner's `unscanned` state are not mapped, and the fixed-severity invariant contradicts the already-accepted conditional rule in ADR 0019. These should be resolved before accepting the ADR. Also, because this changes canonical safety wording, please explicitly note any required public-docs sync per `docs/public-docs-sync.md`/CONTRIBUTING, even if that sync is intentionally deferred.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Confirmed, and thanks for catching this — `validateEntrypoint(...)` (provider entrypoint traversal) and the `setup.files`/`setup.directories` `SafeJoin(...)` check are both real, already-enforced hard-stop checks, and neither is in the 21-row matrix. They're also distinct from row 3 (archive-entry safety, which is pull/import-only): entrypoint traversal is checked at publish/inspect against the resolved provider entrypoint path, and the setup-path check runs at enable/materialize time against `setup.files`/`setup.directories` — different pipeline placement than row 3.

Will add both as explicit rows in a follow-up commit, referencing ADR 0010 and the concrete code paths (`internal/packages/validate.go` for `validateEntrypoint`, the `SafeJoin` call sites for the setup-path check). I'll use the existing lettered-subrow convention (7b, 14a-c) rather than renumbering, so rows 2/13 (PII) don't shift out from under the implementation plan I've just posted below.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Agreed — as worded, "severity is fixed per check, never conditional or escalating" is flatly contradicted two lines later by row 14c, and since 0019 isn't being reopened, the invariant is what has to give, not the shipped behavior.

Will rescope the sentence along these lines: "Severity is fixed per check at the point it's introduced under this ADR — this does not reopen 0019's already-accepted conditional escalation for `disable_safety` (row 14c), which predates and is grandfathered under this policy." That preserves the actual intent (no scoring engine, no per-instance judgment calls for *new* checks) without claiming something row 14c already falsifies.

One thing worth confirming while I'm in there: PII's own sentence right after ("both of its sub-checks are non-blocking...") stands as a specific, standalone commitment for PII, independent of the general invariant — I don't read your comment as touching that, but flagging it explicitly since the PII implementation plan I've posted below relies on it.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Good catch — and this one actually surfaced a real gap in the PII implementation plan I'm posting below, not just a documentation omission. `CategoryUnscanned` is real, shipped, deliberately-separate-from-`WarningFindings` behavior in `instructions.go`, so it belongs in the matrix. Will add it as its own row in a follow-up commit, with `internal/packages/instructions.go` (`CategoryUnscanned`, `UnscannedFindings`) as the owning reference.

Separately: since you're asking for this to be formalized as parent-ADR policy rather than left as an instructions.go implementation detail, I'd suggest writing that row generally enough to cover *any* warning-tier content scanner, not just the instruction-risk one. I'm independently adding the identical "file couldn't be checked" signal for the new PII scanner in response to this exact comment (see Decision 7 in the plan below) — it'd be good for this ADR to name that as a general pattern future scanners should follow, rather than two scanners independently reinventing the same convention with no shared policy line to point to.

Raghavrajvanshi:
Cross-posting the implementation plan for PII detection (issue #129, which this ADR's rows 2/13 are written for) since your review comment above on `CategoryUnscanned` directly changed it — see Decision 7. Full plan, also being posted on #129:

## Proposal: PII & Personal-Context Detection (Issue #129 / ADR 0020 rows 2, 13)

_Updated after this PR's own `CHANGES_REQUESTED` review — Decision 7 below is a direct response to the `CategoryUnscanned` comment on this PR._

### Context

ADR 0020 already settles the policy questions issue #129 raises: PII detection splits into two non-blocking checks — a **Risky**-tier high-risk-pattern check (SSN / credit-card / gov-ID shapes, row 2) and a **Sensitive**-tier general-contact-info check (name / email / phone / address, row 13). Both are `Planned`; neither exists in code yet.

While tracing the existing analogs (`ScanForSecrets`, `ScanForInstructionRisk`), I found that `Import()` (`internal/packages/import.go:55`) calls `Validate()` but only checks `Passed()`/blocking count — it silently discards `Notes` and `InstructionFindings`. I checked how far that gap reaches by reading `runAdd` (cli.go:1023-1137): `lineage add` doesn't rely on `Import`/`Pull`'s return value for risk display at all — it calls `assessEnable` separately, which runs its own fresh `ScanForInstructionRisk`. So `add` already shows instruction-risk warnings correctly, and will pick up PII warnings for free once §7 wires `ScanForPII` into `assessEnable`. The plumbing gap only affects the standalone `lineage package pull`/`import` subcommands.

Checked against current `develop`: no PII scanner exists there, and `docs/safety.md` still states PII detection is "Not implemented in this build."

**Checked PR #228** (closed #128, the direct precedent for this shape of feature). Two things from that review history shape this plan:
- Five review rounds went into `InstructionFinding.Excerpt` redaction alone. Strong precedent for Decision 5 below: no excerpt field for PII at all.
- A coordination bug was caught between `runAdd` and `enableRef` (`showsRiskInfo`/`preConfirmed`). §7 calls this out explicitly so a PII-only finding doesn't fall into the same gap.

**This PR's own review, directly relevant**: your third comment above (missing `CategoryUnscanned` mapping) exposed that this plan's scanner had no equivalent "file couldn't be checked" signal — see Decision 7.

### Decisions

Per the standing decision on issue #129 (2026-08-24): PII/personal-context categories as listed, warning-only (never hard-stop, per #127), publish requires explicit acknowledgment of findings, registry/public surfaces expose category + file-count only.

1. **Scope**: build exactly ADR rows 2 & 13, plus social/portfolio URLs (still a plain regex). Explicitly out of scope, all for the same reason (needs semantic understanding, not pattern matching): resume-structure detection, salary/recruiter/job-search notes, and customer names/confidential project names in package references.
2. **Publish requires explicit acknowledgment of any findings.** `--yes` can't serve as that acknowledgment as the code stands — it runs *before* validation even executes today.
3. **Registry/public surfaces expose category + file-count only.** Never trusted as the receiver's source of truth per this ADR; a declarative summary only.
4. **One `PIIFinding` type**, mirroring `InstructionFinding` (`Path, Category, Severity, Reason`), `Category` enum spanning both tiers plus `pii_unscanned` (Decision 7), `Severity` always `SeverityWarn`. *Note*: your severity-invariant comment above doesn't affect this — PII's own clause ("both sub-checks are non-blocking") is a standalone commitment, not inherited from the general invariant being reworded.
5. **No excerpt field** — `Path + Category + Reason` only. Reinforced by PR #228: getting excerpt redaction safe took five follow-up fixes for instruction-risk alone.
6. **Fix the Pull/Import plumbing gap for the standalone subcommands only.** `add` needs no change — it already gets this via `assessEnable`.
7. **Mirror `CategoryUnscanned` for PII — prompted directly by your comment on this PR.** Instruction-risk keeps "this file couldn't be checked" as its own `SeverityWarn` category, deliberately excluded from `WarningFindings`. This plan's scanner had no equivalent — an oversized/unreadable file would silently vanish, fine for a hard-stop-only scanner (secrets) but not for an all-warning one like PII, where a user needs to tell "clean" from "unchecked." Adding `pii_unscanned` and a matching `PIIUnscannedFindings` filter closes this the same way your comment asks it closed for instruction-risk.

### Implementation

**1. New scanner: `internal/packages/pii.go`** — `PIICategory` enum (8 content categories + `pii_unscanned`), `PIIFinding{Path, Category, Severity, Reason}`, flat documented `piiPatterns` list. Oversized/unreadable/symlinked files produce a `pii_unscanned` finding instead of silent omission, mirroring `instructions.go`'s handling exactly. `WarningFindings` excludes `pii_unscanned`; a new `PIIUnscannedFindings` returns only that category. Name detection has no reliable regex — proposing manifest-declared author/contact fields only (open question below).

**2. Wire into `Validate()`** — `PIIFindings []PIIFinding` on `ValidateReport`, typed and separate, never affects `Passed()`.

**3. Fold into `PortabilityReport`** — appended into `Warnings`, `HasBlockers()` untouched.

**4. Publish acknowledgment gate** — `Publish(dir, cfg, ackPII bool)` gains a param; non-empty `PIIFindings` + `ackPII=false` returns `ErrPIIAcknowledgmentRequired{Findings}` (same sentinel-error convention as `ErrAlreadyImported`). `runPackagePublish` prints findings on that error, then re-prompts interactively or, non-interactively, requires a new `--ack-pii` flag distinct from `--yes` (open question below). Secrets untouched — still an unconditional hard-stop.

**5. Registry payload** — `Publish()` sends a new `pii` query param (`map[category]count`), alongside existing `entrypoints`/`capabilities`. Not in scope: reading it back through `pull`/`inspect`.

**6. Fix Pull/Import plumbing (standalone subcommands only)** — `Import()` returns `(string, ValidateReport, error)`, threaded through `Pull()`. No change to `runAdd`.

**7. Fold into Enable confirmation, including `add`'s pre-prompt display** — `piiWarnings` **and** `piiUnscanned` on `enableAssessment` (mirroring the existing `warnings`/`unscanned` pair per Decision 7), folded into `needsConfirmation()`/`printEnableWarnings`. `runAdd`'s `showsRiskInfo` must check both new fields too, mirroring the exact bug class already caught in PR #228 review.

**8. Inspect + structured output** — `runInspect` calls `ScanForPII`; unscanned-for-PII files land in whatever section `UnscannedFindings` already renders in, not the PII-findings section. Add a `PIIReport`/`toPIIReports` analog to `InstructionRiskReport` in `structured.go`.

**9. Tests** — `pii_test.go` using the existing fixture convention (`t.TempDir()` + `InitPackage` + `mustWrite`):
- `resume-pack` (intentional) → findings present, `Passed() == true`. `private-notes-pack` (accidental) → Risky finding.
- Publish/inspect output distinguishes PII from secrets. Pull/import regression test for the fixed plumbing.
- `TestPublishRequiresPIIAcknowledgment`, `TestPublishSendsPIISummaryOnly`.
- `TestAddShowsPIIOnlyFindingsBeforePrompt` — modeled on PR #228's `TestAddDecliningSetupDoesNotReportReady`.
- `TestScanForPIIReportsUnscannedOversizedFile` — mirrors the equivalent existing instruction-risk unscanned tests.

### Open questions

- **Name detection mechanism**: no regex reliably catches a freeform personal name. Proposing manifest-declared fields only, or dropping it as a standalone category in favor of co-occurrence with email/phone.
- **`Import()` signature**: full `ValidateReport` vs. a smaller `[]string` warnings slice.
- **Non-interactive acknowledgment mechanism**: proposing a distinct `--ack-pii` flag separate from `--yes`.
- **Row-number stability**: this plan cites "ADR rows 2 & 13" — if your first comment's new rows land as numbered rows rather than lettered sub-rows, these could shift. Will confirm final numbers once this PR merges.
- **Should `pii_unscanned`/`CategoryUnscanned` become one general taxonomy concept?** Given your comment asks for this to be formalized as parent-ADR policy, it may be worth writing it generally enough to cover any warning-tier scanner, not just instructions.go, since PII is independently hitting the identical need right now.

### Verification

- `go test ./internal/packages/... ./internal/cli/...`
- Manual: `validate`/`inspect` on a resume-containing package — PII renders distinctly from secrets/instruction-risk.
- Manual: `publish` on a PII-bearing package — refuses without acknowledgment, succeeds once acknowledged.
- Manual: `pull` on a package with PII — warnings now print. `add`/`enable` on a PII-only package — risk section shows before the prompt.
- Manual: `inspect`/`enable` on a package with an oversized file — surfaced as unscanned, not silently dropped, not counted as a PII warning.

docs/decisions/0020-safety-compliance-taxonomy.md
priyam-jain-2002: Direction looks right. Please keep this PR scoped to ADR 0020's taxonomy/completeness work: add the two already-shipped path-safety checks as explicit matrix entries, with their actual pipeline stages and ADR 0010/code references. Using lettered sub-rows is fine if it preserves stable references. Any implementation work discovered while doing this should be tracked separately rather than folded into this docs-only PR.

docs/decisions/0020-safety-compliance-taxonomy.md
priyam-jain-2002: Yes — the PII-specific non-blocking commitment can remain; this comment is only about the parent invariant contradicting ADR 0019. Please make the exception explicit in the ADR without turning it into a broader redesign: new checks introduced under this taxonomy have fixed severity by default, while ADR 0019's already-accepted conditional `disable_safety` behavior remains authoritative. That keeps this PR descriptive of shipped behavior rather than reopening it.

docs/decisions/0020-safety-compliance-taxonomy.md
priyam-jain-2002: Please add the shipped instruction-risk `CategoryUnscanned` outcome to this matrix, but keep the implementation scope separate. I agree there may be a useful general policy concept here — scanners should distinguish “clean” from “not fully scanned” — however generalizing that across future scanners is a follow-up design/implementation concern, especially because #129 is now discovering the same requirement. For this PR, document the existing ADR 0019 behavior and, at most, state the reusable policy principle without prescribing #129's concrete API/types/plumbing. Please raise a dedicated follow-up issue for the cross-scanner `unscanned` convention if one does not already exist, link it from ADR 0020/#129, and let that issue own questions such as shared taxonomy, report shape, CLI rendering/confirmation semantics, and whether scanners need a common interface.

priyam-jain-2002:
Maintainer scope guidance for the PII plan cross-posted here: please keep #289 focused on making ADR 0020 a complete and internally consistent parent taxonomy. The detailed #129 implementation plan is useful context, but decisions such as `--ack-pii`, changing `Import()`/`Pull()` return plumbing, registry PII summaries, name-detection mechanics, and concrete PII report types should remain in #129 (or narrower implementation issues) rather than becoming implied decisions of this ADR PR.

One item deserves its own follow-up because it now spans scanners: the reusable **unscanned / could-not-be-checked** convention. Please raise a dedicated issue if none exists and link it from this ADR and #129. Scope that issue around the policy/contract across scanners — how “unscanned” differs from a positive warning, where it is surfaced, confirmation semantics, and whether report/CLI behavior should be shared. Do not implement that generalized mechanism in #289.

For the next commit on this PR, please limit changes to: (1) map the missing existing path-safety checks, (2) fix the fixed-severity wording so it does not contradict ADR 0019, (3) explicitly map the already-shipped instruction-risk `CategoryUnscanned`, and (4) note/link follow-up work rather than designing it here. Once that commit is up, I’ll re-review the exact taxonomy wording and matrix against current code.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Fixed in commit `1beb740`.

Added two new rows, verified against code before writing anything rather than assumed:

- **Row 4b** — provider entrypoint path traversal. Confirmed `validateEntrypoint`'s `SafeJoin` call (`internal/packages/discovery.go`) is distinct from row 4's symlink check in the same function, and confirmed it's called from **both** `Discover` (`discovery.go:38,41`, used by inspect/enable/run) and `Validate` (`validate.go:68,71`, used by publish/export/import/pull) — so its pipeline-stage list is Publish, Inspect, Pull/Import, Enable, Runtime, the same breadth as row 5.
- **Row 4c** — `setup.files`/`setup.directories` traversal. Confirmed this is actually checked at **two different roots**: `validate.go` (lines 82–89) checks `SafeJoin(dir, ...)` against the package's own directory as an early catch at publish/validate time, and separately `internal/packages/setup.go`'s `PlanSetup`/`ApplySetup` re-checks `SafeJoin(projectRoot, ...)` against the real receiver project root, called from `assessEnable` at enable time. Both are real, so the row lists Publish, Pull/Import, Enable/Materialize as its stages and cites both files.

Both are hard-stop/Unsafe, referencing ADR 0010 alongside the code paths.

One note: the branch picked up a `develop` merge in between (unrelated commits — cursor adapter, package-weight inspection, etc.), but the PR diff is still exactly the two doc files: `git diff upstream/develop...HEAD --stat` shows 111 insertions across `docs/decisions/0020-safety-compliance-taxonomy.md` and `docs/decisions/README.md`, nothing else.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Fixed in commit `1beb740`.

Reworded the invariant to (verbatim from the file now):

> **New checks introduced under this taxonomy have fixed severity by default — this does not reopen ADR 0019.** Instruction-risk scanning already shipped with its own reviewed, accepted conditional rule (row 14c: `disable_safety` starts as Warn and escalates to Block only when co-located with a blocking category in the same file) before this ADR was finalized. That accepted behavior remains authoritative and is adopted here as-is, not redesigned or reconciled into a broader "conditional severity" mechanism — the fixed-per-check default governs checks this ADR itself introduces, and 0019's grandfathered exception is the one already-shipped departure from it.

This scopes the invariant to checks introduced *under this taxonomy going forward*, and explicitly names 0019's `disable_safety` rule as a grandfathered, authoritative exception rather than something this ADR is reconciling into a general "conditional severity" mechanism — matching your suggested wording rather than inventing a new redesign.

Confirming per your reply: PII's own standalone sentence ("both of its sub-checks are non-blocking...") is untouched — that commitment doesn't depend on the general invariant, so rewording the invariant doesn't affect it.

docs/decisions/0020-safety-compliance-taxonomy.md
Raghavrajvanshi: Fixed in commit `1beb740`.

Added row 14d for `CategoryUnscanned`, verified before writing:

- Confirmed it's real, shipped behavior in `internal/packages/instructions.go` — `CategoryUnscanned` is deliberately excluded from `WarningFindings`, with its own `UnscannedFindings` accessor.
- Confirmed it actually gates something, not just a display label: `needsConfirmation` in `internal/cli/cli.go` treats a non-empty `unscanned` list the same as a non-empty `warnings` list when deciding whether `enable` needs a confirmation prompt.

Per your instruction, I documented existing ADR 0019 behavior and stated the reusable principle at the policy level only, without prescribing #129's concrete types/plumbing:

> **a warning-tier content scanner should be able to say "not fully checked" as a category distinct from both "clean" and "found something," and that distinction should reach the same confirmation gate a positive finding does.**

I did *not* design the cross-scanner mechanism. Instead I added a Follow-Up item asking for a dedicated issue to own that (I checked — none exists yet; searched `repo:agentic-lineage/lineage unscanned` and got zero results), scoped exactly as you described: policy/contract across scanners, where it's surfaced, confirmation semantics, whether report/CLI behavior should be shared. That issue isn't filed yet — happy to file it and link it back into the ADR once you confirm you'd like it filed under a specific title/number, or I can leave that to you.

Separately, while verifying this I caught a related overclaim before it shipped: my draft's Follow-Up section originally asserted `docs/safety.md` was "already accurate" for these checks. I actually `grep`-checked it and found it's missing both this row's behavior and row 4c's setup-path check — corrected the claim and added an honest Follow-Up item noting the gap rather than leaving a false statement in the ADR.

Raghavrajvanshi:
Commit `1beb740` (now merged forward with `develop` at `1e161e0` — unrelated commits only, diff against `upstream/develop` is still exactly the two doc files) addresses all four points:

1. Rows 4b/4c for the missing path-safety checks.
2. Fixed-severity wording rescoped to not contradict ADR 0019 (grandfathers row 14c explicitly, doesn't redesign it).
3. Row 14d for `CategoryUnscanned`, documented as existing behavior with the reusable principle stated at the policy level only — no #129 plumbing/types prescribed here.
4. Follow-Up items added: `docs/safety.md` sync gap (checked directly, it's real), and the cross-scanner "unscanned" convention.

Replied in detail on each of the three original inline threads with what was verified and how.

Filed **#312** for the cross-scanner "unscanned" convention as requested, scoped exactly as you described (policy/contract across scanners, not implementation), and linked it from both the ADR's Follow-Up section and from #129.

Ready for re-review whenever convenient.

Issue #288, opened 2026-09-18
Decouple workflow analysis from provider access and execution


priyam-jain-2002:
## Problem statement

While implementing #106, @Rushilkr77 identified a separate architectural gap in the analysis stage introduced by #104: `lineage analyze` is structurally provider-aware, but its live execution path is currently coupled to Anthropic-specific access/configuration.

That coupling should not leak into the behavioral model or the deterministic compiler. The analysis contract should remain stable while provider execution is isolated behind explicit adapters and access modes.

Credit to @Rushilkr77 for surfacing this boundary while working through the analyze -> model -> compile path in #106 / #287.

## Goal

Make **analysis execution provider-neutral without changing behavioral-analysis semantics**.

The invariant should remain:

```text
workspace + evidence
        |
        v
shared analysis contract
        |
        v
explicit provider/access adapter
        |
        v
ParseModel -> Validate -> quality checks
        |
        v
BehavioralModel
        |
        v
lineage compile
```

`lineage compile` must remain deterministic, provider-neutral, and LLM-free.

### 1. Explicit provider selection

Provider selection must be explicit, for example:

```bash
lineage analyze --provider anthropic
lineage analyze --provider openai
lineage analyze --provider openrouter
```

Do **not** infer the provider from an API key. A wrong inference could send workspace/evidence content to the wrong external service.

### 2. Provider adapters with explicit capability/config contracts

Each provider adapter owns its:

- credential requirements
- endpoint/base URL
- model configuration
- request/response translation
- supported capabilities
- provider-specific errors

Providers that happen to expose similar HTTP shapes may share low-level transport helpers, but should remain distinct adapters. An OpenAI-compatible endpoint is not, by itself, evidence of equivalent provider semantics.

### 3. Shared analysis semantics

Changing providers must not silently change the analysis contract.

All provider paths should feed the same downstream behavior:

- same analysis intent/prompt contract
- same evidence payload semantics
- same `model.ParseModel`
- same `model.Validate`
- same quality checks
- same `BehavioralModel` contract

Provider-specific details stay behind the adapter boundary.

### 4. Access modes

Support explicit access modes with clear trade-offs.

#### Manual

No LLM/network call. Allow an author to create/edit the saved `model.json` manually.

#### API

Use the explicitly selected provider and that provider's configured credentials, e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `OPENROUTER_API_KEY`.

[...]

PR #290, opened 2026-09-19
Decouple workflow analysis from provider access (#288)

Rushilkr77:
## Summary

`lineage analyze` only worked with Claude: the endpoint, the `claude-opus-5` model and `ANTHROPIC_API_KEY` were hardcoded in `internal/analysis/claude.go`, next to vendor-neutral logic. This PR splits the neutral core (prompt, evidence payload, caps, secret scan) into `internal/analysis/provider.go` and adds an `Adapter` registry with distinct `anthropic`, `openai` (Chat Completions) and `openrouter` adapters under `internal/analysis/adapters/`.

- No default provider or model. Both resolve flag, then `LINEAGE_ANALYSIS_*` env, then an `analysis:` block in `.lineage/config.yaml`. Never inferred from an API key or from the author's source tool.
- Key comes only from the provider's env var (or `analysis.key_env`). Never a flag, never saved.
- `--endpoint` override, https only unless localhost.
- Egress notice (provider, host, model) before any send; TTY confirm, otherwise `--yes`. CI never prompts.
- `compile` and `internal/compile` untouched; `--fixture` unchanged.

Breaking: `--provider claude` and the built-in model are removed; use `--provider anthropic --model <id>`. CHANGELOG updated.

## Linked Issue

Closes #288

## Package Impact

- [ ] Documentation only
- [ ] Tests only

Affects `lineage analyze` (agent-assisted analysis) and the `.lineage/config.yaml` schema (new optional `analysis:` block). Compiled packages are unchanged.

## Merge Method

## Safety

The existing secret scan still refuses to send a workspace containing credential files. API keys are read from the environment only and never written to config.

## Verification

Relevant test cases:

- `internal/analysis/adapters/adapters_test.go`: per-adapter endpoint, credential env, auth header and response parsing.
- `internal/cli/analyze_test.go`: `parseAnalyzeArgs` for the new flags; `resolveAnalysisProvider` for no provider, unknown provider, missing model, missing key, no provider inferred from a key, http endpoint rejected, localhost allowed, `--yes` in CI, TTY accept and decline, env selection, fixture needing no provider; a check that core files contain no vendor names.
- `internal/analysis/provider_test.go`: ported tests now run through a vendor-free fake adapter; registry lookup; missing key.

```text
go vet ./...
go test ./...
```

## Release Tracking

Complete this section only for PRs targeting `main`. Not applicable: this PR targets `develop`.

## Notes For Reviewers

- #287 has merged; this is rebased onto `develop` and the diff is just this change.
- Not in this PR: agent-CLI mode (reasons in the #288 comment); a no-echo TTY key prompt (needs a terminal dependency that isn't vendored); an init-time prompt to save the provider choice (config block exists, hand-edited for now).
- Follow-up found along the way: #291 (Codex/Cursor source layouts in the inventory classifier).
- Docs: `docs/guides/compiling-existing-workspaces.md` updated. No website/Wiki sync checked yet.


priyam-jain-2002 (changes requested):
Maintainer review — request changes before merge.

Two correctness/safety blockers remain:

1. **Provider overrides inherit another provider's saved profile.** resolveAnalysisProvider resolves provider, model, endpoint, and key env independently. If a workspace has saved OpenAI endpoint/key-env overrides, then --provider anthropic can retain that OpenAI endpoint and credential variable while sending Anthropic wire traffic. That violates the adapter-isolation objective and makes the egress notice misleading. Treat endpoint/key-env (and ideally model) as a provider-scoped profile: an explicit provider override must not inherit saved values from a different provider. Please add CLI and environment-precedence tests for this case.

2. **The outbound evidence payload is only capped per file, not in aggregate.** Inventory has no entry/total-byte limit; buildSourceExcerpts admits every file up to 8 KiB, and then the whole inventory plus every excerpt is marshalled and sent. A workspace containing many small files can therefore produce unbounded request size, cost, latency, and data egress despite the documented caps. Enforce a deterministic total request budget before marshaling/sending, with a test using many small files. Prefer an explicit actionable refusal when the evidence exceeds the budget, or clearly documented deterministic inclusion behavior.

The adapter split, explicit provider selection, shared Parse/Validate path, endpoint safety policy, consent flow, documentation, and green CI are otherwise headed in the right direction. I do not consider the deferred agent-CLI adapter a blocker for this API/manual PR, but it needs a dedicated tracked follow-up: the current Adapter/RemoteProvider contract requires CredentialEnv and an API key, so a CLI/subscription mode cannot be added without introducing an access-mode seam rather than merely another adapter.

priyam-jain-2002 (commented):
Additional required fix:

3. **Resolve project analysis configuration from the analysis target, not the process working directory.** `lineage analyze <path>` inventories and sends evidence for `<path>`, but the current code discovers `.lineage/config.yaml` from `os.Getwd()`. For example, running `lineage analyze ~/projects/customer-app` while standing in `~/projects/lineage` applies Lineage's provider/model/endpoint configuration to `customer-app`. Resolve config by walking upward from the normalized analysis target path instead. Preserve flags and environment variables as higher-precedence overrides, and add a regression test with different CWD and target-workspace configurations that proves the target configuration wins.

Rushilkr77:
Thanks for the review. All three required fixes are pushed (`45d8efd`, plus a follow-up test commit), and the agent-CLI follow-up is now tracked.

**1. Provider-scoped profile.** Provider is resolved first, then every other setting is scoped to it. Saved `model` / `endpoint` / `key_env` apply only when `analysis.provider` equals the selected provider. Env `LINEAGE_ANALYSIS_MODEL` / `LINEAGE_ANALYSIS_ENDPOINT` apply only when `LINEAGE_ANALYSIS_PROVIDER` is unset or equal to the selected provider. So `--provider anthropic` no longer inherits a saved OpenAI endpoint, model or key variable, and the egress notice shows the real host. Tests cover saved-config and env precedence for anthropic, openai and openrouter, in both directions.

**2. Total evidence budget.** `MaxEvidenceBytes` = 512 KiB for the whole outbound payload (inventory plus excerpts, as JSON). It is enforced while building excerpts and again after marshaling. Over budget, the run refuses with an actionable error and sends nothing; nothing is silently trimmed. A test with many small files passes every per-file cap, trips the budget, and asserts no request was made. The guide documents the limit.

**3. Config from the analysis target.** Project config is now found by walking up from the normalized `<path>`, not the working directory. A target with no config does not fall back to the CWD's. Flags and env still take precedence. The regression test uses different config in the CWD and the target and asserts the target wins.

**Agent-CLI follow-up:** #292. You are right that the current `Adapter` / `RemoteProvider` contract (`CredentialEnv`, API key) means a CLI mode needs an access-mode seam rather than another adapter; the issue says so and scopes it: explicit mode menu and consent, non-interactive single-shot session, isolation from workspace and user-level instruction files, and how the output is verified. My earlier comment on #288 said no core changes would be needed; that was wrong.

`go vet ./...` and `go test ./...` pass locally. I'm leaving the PR as a draft for now.

priyam-jain-2002 (commented):
Follow-up on blocker 2: this is close, but not fully addressed yet. The source-excerpt loop is capped before serialization, but json.Marshal(evidencePayload{Inventory: inv, ...}) still serializes the entire unbounded inventory before len(evidence) > MaxEvidenceBytes rejects it. A workspace with very many empty/tiny files can therefore force unbounded inventory allocation/serialization even though nothing is ultimately sent.

Please account for the inventory metadata in the deterministic budget before marshaling the complete payload (or impose an explicit inventory entry/serialized-metadata limit before constructing it), then retain the final serialized-size assertion as defense in depth. Add a test with enough empty files to exceed the inventory-only budget and assert no adapter request is made. The target-config and provider-profile fixes look correct.

Rushilkr77:
Thanks, agreed: the excerpt loop was capped but `json.Marshal` still walked the whole inventory first. Fixed in `c1d39cd`.

- `inventoryBytes(inv)` estimates the inventory's serialized size from its string fields (paths, digests, reasons, and every mention / referenced-by citation) plus a fixed per-object overhead, without marshaling.
- `buildSourceExcerpts` starts its running total from that estimate and refuses over `MaxEvidenceBytes` before reading any file or building the payload. So the inventory metadata and the excerpts spend one shared budget, deterministically.
- The final `len(evidence) > MaxEvidenceBytes` check after marshaling stays as defense in depth (JSON escaping can add a little on top of the estimate).
- New test: enough empty files that the inventory metadata alone exceeds the budget (excerpts are zero bytes, so only the inventory budget can trip). It asserts the refusal and that no adapter request was made.
- The guide now says the limit counts inventory metadata as well as file contents.

One thing this does not bound: `inventory.Discover` itself still builds the full inventory before analysis starts, since the caller runs it. That is existing behavior and out of this PR's scope; the send path is now bounded.

`go vet ./...` and `go test ./...` pass.

Rushilkr77:
Follow-up on the remaining gap I flagged in my last comment (`Discover` building the full inventory before any budget applied). Fixed in `9cdfb61`, so each stage now has its own cap, earliest first:

1. **Walk:** `inventory.DiscoverWithLimit(root, n)` refuses with `ErrTooManyFiles` as soon as the walk admits more than `analysis.MaxInventoryEntries` (2000) regular files, before any file is hashed or scanned for citations. `analyze` uses it only when a live provider will receive the workspace; `--fixture` sends nothing, so it keeps the unbounded inventory, and `compile` still calls plain `Discover` (unchanged).
2. **Inventory metadata:** `inventoryBytes` estimate against `MaxEvidenceBytes` (previous commit), which also covers citation volume.
3. **Excerpts:** running total against the same budget while reading files.
4. **Final:** `len(evidence)` after marshaling, as defense in depth.

Also `digestFile` now streams through `io.Copy` into the hash instead of `os.ReadFile`, so hashing one very large file no longer loads it into memory (shared code, same digests).

New test: `DiscoverWithLimit` at, under and at zero (unbounded) the file count. The guide documents the 2000-file, 512 KiB limits. `go vet ./...` and `go test ./...` pass.

priyam-jain-2002 (approved):
Re-review complete. The remaining aggregate-evidence concern is addressed: live analysis now bounds inventory during discovery before hashing/citation scanning, accounts for inventory metadata before evidence serialization, retains the final serialized-size guard, and tests both source-content and inventory-only overflow paths. The earlier provider-profile isolation and target-workspace configuration fixes remain correct. Approved.

